PR #2, opened 2016-05-01
Fix disabling SAML login when duallogin is off

keesfransen:
Hi Brendan,

I'm a fairly new to Moodle dev so if you think there is a better way of catching this then let me know.

Thanks for the great plugin!

Cheers,
Kees

### DIFF
--- a/auth.php
+++ b/auth.php
@@ -134,6 +134,7 @@ class auth_plugin_saml2 extends auth_plugin_base {
      * All the checking happens before the login page in this hook
      */
     public function loginpage_hook() {
+        global $SESSION;
 
         $this->log(__FUNCTION__ . ' enter');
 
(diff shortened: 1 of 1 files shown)


brendanheywood:
Thanks heaps for the patch @keesfransen  :)

There are a couple minor issues and then I'll be happy to pull this in:

1) travis is failing on a couple minor code standards things, can you please fix those and force push a squashed commit? eg

https://travis-ci.org/CatalystIT-AU/moodle-auth_saml2/jobs/126988896

2) I find $SESSION->saml a little bit vague, maybe rename it to ->saml_skipped and on line 163 set it to 1, and then on line 155 simply say

if (!empty($SESSION->saml_skipped ) {

Also I think this will introduce a regression bug: there is a small edge case situation where you could come in via a link with ?saml=0, but then not actually login and keep browsing as a guest, and then later click on a link which has ?saml=1 which would *should* force you to the idp, but won't. I think the solution is to unset($SESSION->saml_skipped) on line 166 if $saml != 0, and also you'll need to swap the order so your new if test comes after the setting of $SESSION->saml_skipped, otherwise it will never run. But it should only be unset if the query param is explictely ?saml=1 and not simple defaulting to this on line 161. This needs some careful testing, it would be good if you could add some testing steps to this PR to reproduce both the original bug you fixed and for this second edge case

keesfransen:
Hey Brendan,

Thanks for the feedback. Sorry I didn't notice the travis check when I opened the PR.

Do you think 
```
if ($saml || isguestuser()) {
    unset($SESSION->saml_skipped);
}
```
after the dual login check would work ok? I think this would cover cases where the query param was explicitly set on or if the user became a guest (via `?saml=0` or `$CFG->autologinguests`).

Will get those code issues fixed and testing steps done.

Cheers,
Kees

brendanheywood:
I wouldn't use isguestuser() as they could be anon as well which is subtly different. But either way I'd start by getting a test script in place which produces the issue and only then start tweaking the code to fix it.

Also there is another issue unrelated to yours which is breaking the travis build, I'll fix that shortly please ignore any issues with the db/install.xml file

brendanheywood:
Thanks @keesfransen I've pulled this in patch there were still a few more edge cases bug this introduced which I've now also fixed, ie if you try to login with a forced ?saml=off url, fail the authentication then you correctly get the manual login page again. But then if you don't login and goto some other completely random deep link which requires auth then you will correctly get booted to the IdP.

keesfransen:
Hi Brendan. My apologies you never heard back from me, this ended up on my backlog 😢 . Thanks for bringing this to completion!

brendanheywood:
No worries, thanks for the patch :)

PR #19, opened 2016-07-14
configure SAML plugin to sign logout requests by default.

cristianomariacumer:
Configure SAML plugin to sign logout requests by default's needed by some IdP like ADFS


brendanheywood:
Thanks!

Issue #23, opened 2016-07-26
User profile field locking is not respected


mattporritt:
When user profile fields are locked in the SAML2 auth configuration page, users are still able to edit the locked fields in their own profile. This is present in Moodle 2.9 (but looks like it would affect other versions)
Dirty test script:
* Install and configure SAML2 auth plugin
* Configure firstname to be locked in saml2 auth settings
* Configure a non admin user to have saml2 auth
* Login in as the saml2 user
* Edit user profile

Firstname field is still editable.
This looks to be an incarnation of this tracker issue: https://tracker.moodle.org/browse/MDL-36987. I have a patch in the works that I'll lodge as a pull request once it is tested

brendanheywood:
Correct fix should just use the get_config('auth/saml2') and have an upgrade script to fix legacy config data

PR #32, opened 2016-07-29
Issue#23 field locking - Redux

mattporritt:
Initial fix for this issue broke idp integration as the array merge overwrote required values.
I've tested this new fix against a live system and everything behaves as it should.

I'm still not happy with the way this problem is solved. I had a serious go of trying to do it a better way by implementing it at the configuration stages. But authlib is old school.  I couldn't figure out how to do it with refactoring lots of core authlib.

### DIFF
--- a/db/upgrade.php
+++ b/db/upgrade.php
@@ -74,20 +74,19 @@ function xmldb_auth_saml2_upgrade($oldversion) {
 
     if ($oldversion < 2016080302) {
         // Update plugin configuration settings from auth_saml2 to auth/saml2.
-        $current_config = get_config('auth_saml2');
-
+        $currentconfig = get_config('auth_saml2');
 
         // Remove old config.
         $rs = $DB->get_recordset_select('config_plugins', 'plugin = ?', array('auth_saml2'));
         foreach ($rs as $record) {
-            if ($record->name != 'version'){
+            if ($record->name != 'version') {
                 $DB->delete_records('config_plugins', array('id' => $record->id));
             }
         }
         $rs->close();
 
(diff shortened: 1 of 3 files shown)


brendanheywood:
I reopened #23, can you please instead rename the config from 'auth_saml2' to 'auth/saml2' and add an upgrade script to move them to where they should be?

db/upgrade.php
```diff
@@ -72,6 +72,18 @@ function xmldb_auth_saml2_upgrade($oldversion) {
         upgrade_plugin_savepoint(true, 2016031701, 'auth', 'saml2');
     }
 
+    if ($oldversion < 2016071201) {
+        // Update plugin configuration settings from auth_saml2 to auth/saml2
+        $rs = $DB->get_recordset_select('config_plugins', 'plugin = ?', array('auth_saml2'));
```
brendanheywood: There is a bug with this with the 'version', because core stores it as:

59│ 1272 ║ auth_saml2 │ version                             │ 2016071200

We need to leave it alone

db/upgrade.php
```diff
@@ -72,6 +72,18 @@ function xmldb_auth_saml2_upgrade($oldversion) {
...
+        foreach ($rs as $record) {
+            $record->plugin = 'auth/saml2';
+            $DB->update_record('config_plugins', $record);
```
brendanheywood: Maybe I'm being a bit paranoid, but I think we should $DB->delete_record() on the old data, and then set_config($key, $config->$key, 'auth/saml2'); on the new

PR #37, opened 2016-08-02
Support for Moodle custom profile field mapping

jlyle29:
I have added code to enable mapping of custom moodle profile fields.

Could do with a point in the right direction in terms of custom profile field validation because of the following. Currently the custom fields are saved back without validation therefore if the data is not correct an error is displayed to the user and they cannot log in. eg custom field is a drop down with accepted values 1,2,3,4 if the request gets value 5 an error is given.

This is the first time I have used GitHub, and my first time coding for Moodle so a few things may be bad practice. This is also a copy and paste from my code rather than a pull into my repo so apologies if I have forgotten to copy something over.

### DIFF
--- a/auth.php
+++ b/auth.php
@@ -264,7 +264,7 @@ class auth_plugin_saml2 extends auth_plugin_base {
                     $updateonlogin = $mapconfig->{'field_updatelocal_'.$field} === 'onlogin';
 
                     if ($newuser || $updateonlogin) {
-                        //Determine which fields are custom profile fields and act accordingly
+                        // Determine which fields are custom profile fields and act accordingly.
                         if (in_array($field, $customFields)) {
                                 $user->{'profile_field_'.$field} = $attributes[$attr][0];
                                 //TODO WARNING data validation should be done before saving custom fields back currently if the SAML data is not of the correct type an error is displayed to the user and they cannot log in!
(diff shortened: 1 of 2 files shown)


brendanheywood:
thanks @jlyle29 :)

Re validation, have a poke around here in the user profile lib:

https://github.com/moodle/moodle/blob/master/user/profile/lib.php#L474

You may be able to call that api directly with a bit of massaging, it is designed to be called with form data. I'm assuming that if the data doesn't validate then you'd just discard it and retain the existing data? Also be aware that some user field types don't store the data as-is, they might map it, or escape it or whatever. So call edit_save_data_preprocess() on the respective field class to transform it ready for saving, eg:

https://github.com/moodle/moodle/blob/master/user/profile/field/menu/field.class.php#L119

Also some minor house keeping stuff, can you please update your code to conform to the moodle code standards, you can look here to see the travis results (or in the pull request page):

https://travis-ci.org/CatalystIT-AU/moodle-auth_saml2/jobs/149130708#L297

https://docs.moodle.org/dev/Coding_style

thanks again!

jlyle29:
Awesome thanks for the pointers, hoping to get a chance to follow them up soon.

brendanheywood:
Sorry @jlyle29 for forgetting this, was this ready for a review and integration? Are you using this locally?

nhoobin:
Thanks for this patch @jlyle29.

I have modified and tested out your code, custom profile fields can now be saved and updated.

The new PR is #49, and I will close this shortly.

PR #49, opened 2016-10-12
Support for Moodle custom profile field mapping - part two.

nhoobin:
This builds upon and implements pull request #37 Support for Moodle custom profile field mapping.

As of yet there is no content validation of the fields. If the field mapping is invalid/missing then the data will not be updated.

### DIFF
--- a/auth.php
+++ b/auth.php
@@ -253,12 +253,7 @@ class auth_plugin_saml2 extends auth_plugin_base {
         }
 
         // Grab custom profile fields list for use later on
-        // TODO Probably a better way to do this?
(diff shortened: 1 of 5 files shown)


nhoobin:
I have reduced the complexity of the profile field checking in the latest commit 5a1ff3a.

brendanheywood:
saml_login() is pretty big, can you split that big chunk of attribute logic off from it from about line 278 (30 lines) into a new function that returns a list of attribitues that gets merged on top of the user object and then wrap some unit tests + dataprovider around it please

Issue #51, opened 2016-10-13
SAML1 (shib13) not implemented properly


eSrem:
One of my clients uses a SAML1 IdP and thus returns shib13 in their metadata. I analysed a problem and found what I can change to make it work. It needs a rework of two files. I will fork the repository and do a pull request.

eSrem:
It would be awesome if you could please merge #52

PR #52, opened 2016-10-13
Issue #51 SAML1 shib13 implemented

eSrem:
The SP module did not create the shire link properly.
The SP metadata does not have separate ACS for Artifact.
I took it upon myself to remove the Artifact code from the SSO1 method.


eSrem:
I forgot a global $CFG... i will recommit.

PR #53, opened 2016-10-13
Issue 51: Shib13 not implemented properly

eSrem:
shire link is now set correctly. Removed artifact code. Ensured CFG is accessible from the space.

### DIFF
--- a/extlib/simplesamlphp/modules/saml/lib/Auth/Source/SP.php
+++ b/extlib/simplesamlphp/modules/saml/lib/Auth/Source/SP.php
@@ -153,16 +153,7 @@ class sspmod_saml_Auth_Source_SP extends SimpleSAML_Auth_Source {
 		$id = SimpleSAML_Auth_State::saveState($state, 'saml:sp:sso');
 		$ar->setRelayState($id);
 
-		$useArtifact = $idpMetadata->getBoolean('saml1.useartifact', NULL);
-		if ($useArtifact === NULL) {
-			$useArtifact = $this->metadata->getBoolean('saml1.useartifact', FALSE);
-		}
-
-		if ($useArtifact) {
-			$shire = SimpleSAML_Module::getModuleURL('saml/sp/saml1-acs.php/' . $this->authId . '/artifact');
-		} else {
-			$shire = SimpleSAML_Module::getModuleURL('saml/sp/saml1-acs.php/' . $this->authId);
-		}
+        $shire =  $CFG->wwwroot . '/auth/saml2/sp/saml1-acs.php/' . $this->authId;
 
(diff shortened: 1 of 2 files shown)


brendanheywood:
hi @eSrem thanks for the PR! :)

Saml1 support is a good thing, the only thing I'm a little uneasy about in this patch is that you are modifying the 3rd party libs. Wherever possible I've tried to avoid these and keep them clean so we can more easily upgrade. You should be able to achieve the same thing by modifying /sp/metadata.php either by adding the saml1 elements to the xml or by modifying the xml produced on the fly.

eSrem:
Hi Brendan, I understand you being uneasy. I agree on being able to 'drop in' the external libs as needed. However changes have been made to the lib to make the drop in lib work:
https://github.com/catalyst/moodle-auth_saml2/blob/master/extlib/simplesamlphp/modules/saml/lib/Auth/Source/SP.php#L193

SimpleSAML_Module::getModuleURL will ask for a BaseURL which will be $CFG->wwwroot . /auth/saml2/sp, however the method call append 'module.php' and then /saml/sp/saml2-acs.php/. Making the full AssertionConsumerURL something like:

- $CFG->wwwroot/simplesaml/module.php/saml/sp/saml1-acs.php/$authid
- $CFG->wwwroot/simplesaml/module.php/saml/sp/saml2-acs.php/$authid

We can change the /simplesaml part after the wwwroot from the config.php and setting a baseurlpath. However the resulting URIs will be:
- $CFG->wwwroot/auth/saml2/sp/module.php/saml/sp/saml1-acs.php/$authid
- $CFG->wwwroot/auth/saml2/sp/module.php/saml/sp/saml2-acs.php/$authid

This is if we do not want _any_ changes in the external libs and be able to drop in simplesamlphp as we go. (Which we should)

For this we'd need to:
- add a module.php in the sp directory which magically calls the "/saml/sp/saml2-acs.php/$authid" part.
- undo all the changes made to the supplied simplesamlphp so we can succesfully drop in versions as required (there might be more, i don't know about them yet :D)
- change metadata generation so that the ACS are properly reflected as the second link type.

This change will improve upgradability of the lib. However it will affect all installations as the ACS endpoints will change, and therefore the SP metadata will change. And every single plugin instance will need to reconfigure their IdP's with the new metadata. This is a trade-off.

What it comes down to is: do we want to develop cleanly and let the external lib be an external lib, and therefore have the users of the plug-in change their configurations? Or do we hack the external lib like it was done in the SSO2 method?

I want to hear your €0,02!

brendanheywood:
thanks @eSrem - what you suggested is a better way, and I should have done that originally. But breaking existing sp metadata is a no go for me so I will merge this as-is. thanks again :)

Issue #39, opened 2016-08-05
auto-create users fails when mapped attribute is null


jlyle29:
I have an ADFS 2.0 IDP and have an Active Directory (AD) attribute mapped to the Department Moodle profile field.

It appears that if the AD attribute is not set the claim is not passed at all (needs confirmation). When user auto creation is on a blank profile is created but the profile fields fail to become set because the key in the auth.php $attributes var does not exist.

Therefore Moodle sql error is given  stating that Department cannot be NULL.

gradyumiller:
For any attribute set to update local "On every login" if the value is null for a specific user, he or she cannot login. An error is displayed: Error writing to database. 

It would great if this worked just like it does for LDAP authentication, where users can login even if some of linked the attributes are null.

nhoobin:
Thanks for the earlier commits @jlyle29. It should no longer throw an exception about a null constraint writing to the database.

PR #54, opened 2016-10-16
Supplying tests for refactored update_user_profile_fields.

nhoobin:
Adding tests for pull request #49 Support for Moodle custom profile field mapping - part two.

Refactored saml_login() and extracted a new method update_user_profile_fields().


nhoobin:
This contains a fix for issue #39 auto-create users fails when mapped attribute is null.

PR #56, opened 2016-10-20
Added configuration option to disable the signature in the SP metadata.

nhoobin:


lang/en/auth_saml2.php
```diff
@@ -39,6 +39,8 @@
 $string['spmetadata'] = 'SP Metdata';
 $string['spmetadata_help'] = '<a href=\'{$a}\'>View Service Provider Metadata</a> | <a href=\'{$a}?download=1\'>Download SP Metadata</a>
 <p>You may need to give this to the IdP admin to whitelist you.</p>';
+$string['spmetadatasign'] = 'SP Metdata signature';
+$string['spmetadatasign_help'] = 'Signature pinning for the SP Metadata';
```
brendanheywood: pinning ? is that a typo?

PR #63, opened 2016-11-24
Correct logout link for this Moodle installation

throup:
If an SAML assertion is received but it cannot be matched against a Moodle user, an error is displayed with a Logout link.

auth_saml2 error

This link is relative to the server root, not to the root of the Moodle installation. This is not a problem if Moodle is installed directly into the server root, but does not work if it is installed in a subdirectory.

For example, if Moodle is installed at:

https://www.example.com/subdomain/

then the link should be:

https://www.example.com/subdomain/auth/saml2/logout.php

but it will actually point to:

https://www.example.com/auth/saml2/logout.php

### DIFF
--- a/auth.php
+++ b/auth.php
@@ -150,11 +150,13 @@ class auth_plugin_saml2 extends auth_plugin_base {
     public function error_page($msg) {
         global $PAGE, $OUTPUT, $SITE;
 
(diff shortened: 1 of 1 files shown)


brendanheywood:
thanks @throup :)

PR #64, opened 2016-12-01
Implementation of auth hook, ignore_timeout_hook.

nhoobin:
This hook is now patched in upstream stable 30, 31, master.
https://tracker.moodle.org/browse/MDL-56417


brendanheywood:
Can you please cover this with some tests? Not sure how deep we can test properly, but we should at least be able to setup some users with old session data in various states and run the cleanup task and confirm they are or are not removed

PR #83, opened 2017-02-07
Handle OPENSSL_CONF env var if defined

LukeCarrier:
The OpenSSL extension doesn't always correctly handle the `OPENSSL_CONF` environment variable. We should be explicit about which configuration file we'd like to use if the user has declared it within their environment.


brendanheywood:
@LukeCarrier this looks good too, there was a merge conflict with the other windows stuff but I've fixed that and merged it in. If you could also make a PR updating the README.md with some words around the env var that would be cool too

PR #84, opened 2017-02-09
Document OPENSSL_CONF env var

LukeCarrier:
Provides documentation for the `OPENSSL_CONF` environment variable as suggested in #83.

I've placed it under the gotchas section of the README because it shouldn't be relevant to users of standard configurations, but am happy to relocate it. I've linked to the relevant OpenSSL documentation too.

Thanks for the prompt responses to PRs, much appreciated!


brendanheywood:
Nice one thanks :)

PR #85, opened 2017-02-09
Fix path handling on Windows

LukeCarrier:
Under Windows, it appears as though paths containing drive letters aren't handled correctly in `SimpleSAML_Configuration::resolvePath()`. They should be treated as absolute, but SimpleSAMLphp only treats paths starting with `/` this way.

I realise I've patched SimpleSAMLphp to achieve this and I can see that there are some Moodle-specific patches in SimpleSAMLphp but couldn't find any documentation on these. I suspect this patch really belongs upstream, but I would appreciate it if we could include a workaround in the plugin's fork for the time being?

Thanks in advance.

### DIFF
--- a/extlib/simplesamlphp/lib/SimpleSAML/Utils/System.php
+++ b/extlib/simplesamlphp/lib/SimpleSAML/Utils/System.php
@@ -114,6 +114,10 @@ class System
             $base = $config->getBaseDir();
         }
 
+        // normalise directory separator
+        $base = str_replace(DIRECTORY_SEPARATOR, '/', $base);
+        $path = str_replace(DIRECTORY_SEPARATOR, '/', $path);
+
         // remove trailing slashes
         $base = rtrim($base, '/');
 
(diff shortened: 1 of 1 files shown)


brendanheywood:
Hmm, this one seems more like a bug in simplesaml itself. I think it may be better if we could raise an issue + fork + fix there. It would also need some test coverage here:

https://github.com/simplesamlphp/simplesamlphp/blob/6f0ca916eed1f2a3d0de2f31162edeef386d959f/tests/lib/SimpleSAML/ConfigurationTest.php#L193-L204

I've just made an issue and linked back to here:

https://github.com/simplesamlphp/simplesamlphp/issues/559

brendanheywood:
@LukeCarrier do you have any appetite to try and get this resolved upstream in the simplesaml repo? This is a simple looking change but I just have no capacity to test this myself on windows

LukeCarrier:
@brendanheywood That's the plan, I'm tied up with other things in my day job at the moment but should get to this before EOD Friday.

LukeCarrier:
@brendanheywood thanks for the input; I've opened PR simplesamlphp/simplesamlphp#564 for this change now. Feel free to close if you'd rather not backport something.

brendanheywood:
awesome, can you ping me or raise an issue when SSP has integrated that into the latest stable and I'll upgrade to it

PR #92, opened 2017-02-21
Add hook for custom parameters

LukeCarrier:
I'm not convinced this is the tidiest way to accomplish the intended result, so feedback is most definitely welcome. If some form of middleware implementation is planned for this plugin, this might be an interesting requirement to accommodate through some other means.

TL;DR: adds a hook to `\auth_plugin_saml2::saml_login()` which allows you to specify custom `$params` to `\SimpleSAML_Auth_Simple::requireAuth()`.

The B2B portion of our business had a requirement for a SAML2 configuration between multiple SPs (all Moodle at present, though this may change in the future) and multiple IdPs. In order to achieve this, we configured a standalone SimpleSAMLphp instance to behave as an identity broker, behaving an SP to external IdPs and providing an internal IdP using its own SAML SP as an authentication source.

In this configuration, following links to the identity broker to authenticate will prompt SimpleSAMLphp to display a list of available identity providers, allowing the user to select the one they wish to authenticate with.

In order to facilitate white labeling a Moodle instance, we configure multiple domains for the same Moodle installation and provide different theme configurations based on the domain. This patch enables us to also customise the SAML2 plugin's configuration, appending different [scoping options](https://simplesamlphp.org/docs/stable/simplesamlphp-scoping) to the `AuthnRequest`, e.g.:

```php
$CFG->auth_saml2_params = function() {
    return array(
        'saml:IDPList' => array(
            'http://example.com/saml2/idp/metadata.php'
        ),
        'saml:ProxyCount' => 2,
    );
};
```

### DIFF
--- a/auth.php
+++ b/auth.php
@@ -279,8 +279,15 @@ class auth_plugin_saml2 extends auth_plugin_base {
 
         require_once('setup.php');
         require_once("$CFG->dirroot/login/lib.php");
+
+        if (!empty($CFG->auth_saml2_params)) {
+            $params = call_user_func($CFG->auth_saml2_params);
+        } else {
+            $params = array();
+        }
+
         $auth = new SimpleSAML_Auth_Simple($this->spname);
-        $auth->requireAuth();
+        $auth->requireAuth($params);
         $attributes = $auth->getAttributes();
 
         $attr = $this->config->idpattr;


brendanheywood:
Yeah that's a tricky one to get right. If it was just values then we could turn that into config items. But functions means you are writing code somewhere so it's no longer in the realm of simple config items. At the same time I don't want to simply expose the entire inner working and config of SSP in the plugin settings.

If this was us internally, we simply fork the plugin and keep the code in the fork. Our continuous integration automatically merges with the upstream canonical so we get updates for free and keep the client specific code in branches or completely new private repos.

So sometimes git is just the answer instead of code

LukeCarrier:
I notice in #5 that there is interest in (and possible sponsorship for) implementing support for multiple identity providers directly in Moodle. Is dynamic selection of the identity provider likely to be a requirement of this work? I'm wondering if there's another way to introduce this feature that doesn't involve directly exposing the internals.

brendanheywood:
When #5 lands it will mean on the moodle login page you will get N login buttons, 1 for each idp. It would also mean the fast redirect directly to the idp can't happen. So the student is choosing.

In your case however I think that all you actually want is moodle 1 to use idp1 and moodle 2 to use idp2 etc?? If that's the only thing you need then perhaps a better way is just some logic in your config.php which dynamically sets the idp url via $CFG->forced_plugin_settings for the saml plugin. The only missing piece of the puzzle is that when you normally set that in the gui, it downloads the idp xml and stashes it into sitedir/saml2/idp.xml  So what you would need is for the filename to not be hard coded and then you can have as many idp files in there as you want, and the config just points to the right version and/or grabs it on demand.

I don't think you'd actually want the 'proper' multi auth in #5, because then you'd be able to login to moodle 1 from idp 2, which I assume isn't what you'd want. You lock this down in the idp but the student would still get the choice on the login page.

The sp metadata that moodle generates for itself is actually already like this. This was by design so that you can have the reverse situation which is to have dev / uat / prod moodles all with auto generated certs and metadata and store them all in prod so that after a refresh all the other envs still work. eg

```sh
/var/lib/sitedata/moodle.local/saml2 $ ls -l
-rw-rw-rw- 1 www-data www-data 3549 Dec 20 09:48 idp.xml
-rw-rw-rw- 1 www-data www-data 1472 Jul 12  2016 moodle.local.crt
-rw-rw-rw- 1 www-data www-data 1834 Jul 12  2016 moodle.local.pem
-rw-rw-rw- 1 www-data www-data 1887 Dec 20 10:44 moodle.prod.local.crt
-rw-rw-rw- 1 www-data www-data 1834 Dec 20 10:44 moodle.prod.local.pem
-rw-rw-rw- 1 www-data www-data 8869 Feb 14 09:44 moodle.prod.local.xml
-rw-rw-rw- 1 www-data www-data 1472 Aug  8  2016 moodle.stage.local.crt
-rw-rw-rw- 1 www-data www-data 1834 Aug  8  2016 moodle.stage.local.pem
-rw-rw-rw- 1 www-data www-data 1472 Aug  8  2016 moodle.test.local.crt
-rw-rw-rw- 1 www-data www-data 1834 Aug  8  2016 moodle.test.local.pem
```
 
So if all that makes sense and would solve it then I'd happily accept a pr something like:

* [ ] refactor all the logic of creating the filename for the idp metadata based on the idp url. ie instead of idp.xml if your idp was google.com/saml/ then the file might be sitedata/saml2/idp.google.com.saml.xml It could even just be the full url but escaped. I think it needs the prefix or be in a subdir to protect against wiping out any sp metadata overlap.
* [ ] when reading this file, check if the legacy idp.xml file exists and use it
* [ ] on save in the gui, delete the legacy file and make the new one

This would also mean the whole saml handshake would only be done once and be a whole lot faster without the middle SSP broker instance slowing things down. One of the main reasons I start writing this plugin was because I wanted to optimize the login time as much as possible as it's always the bottle neck of the perceived first moodle page render.

LukeCarrier:
@brendanheywood apologies for the radio silence!

Totally correct, we want to avoid exposing the complete list of IdPs to users. Whilst I appreciate the performance point, the IdP Proxy is a necessary evil in allowing us to introduce additional platforms around Moodle (we'd rather that users do not have to SSO in to Moodle only to immediately leave for another platform). We already have all of the components in place to alter the site configuration based upon the domain, so exposing `IDPList` as a configuration property or via a callback is enough for us to direct users to the appropriate IdP.

Do you know what the configuration schema looks like for #5 yet? I am happy to rework this patch to expose `IDPList` as a configuration parameter if it'd be accepted, but also understand if you'd rather we fork.

brendanheywood:
@LukeCarrier you can see the multi branch here:

https://github.com/catalyst/moodle-auth_saml2/tree/MOODLE_33PLUS-multiidp

It's still waiting on client feedback and will be merged in at some point

LukeCarrier:
There's no obvious way to add configuration options for the entire `$params` option as there are tens of keys used throughout `SimpleSAML_Auth_Source::doLogin()`.

I'm not sure how to rework this patch. As I see it, the options are:

1. Expose just the `IDPList` option as a configuration option accessible via the UI. This feels like a clumsy workaround, as it's visual noise for most users configuring an already complex component and it doesn't allow a great degree of customisation by developers. It would allow us to set the value via `$CFG->forced_plugin_settings` as described above, though.
2. Remove the current params hook, instead allowing developers to set e.g. `$CFG->auth_saml2_params`. I fear this isn't a great improvement on the existing patch.
3. Maintain a fork. I'd rather avoid this, but I appreciate that it complicates maintenance for the sake of a fairly small user base.

@brendanheywood would you accept one of options 1 or 2? If not, we can close this PR.

Thanks in advance! :smiley:

brendanheywood:
@LukeCarrier I'm still not perfectly sure I understand your requirements, but I think I'm leaning towards you turning saml:IDPList into a new config item. When you say 'visual noise' are you meaning for admins configurint this or for the users? My understanding is that the end users won't see anything different here. If you feel this is fairly advanced config then make it a config.php only setting but make sure it is documented in the README here:

https://github.com/catalyst/moodle-auth_saml2#configuration

If you can please make 2 PR's for the 2 stable branches.

Issue #62, opened 2016-11-16
Simple saml lib is on unstable trunk and should be on a stable eg 1.14


brendanheywood:
* [x] upgrade to 1.14
* [x] update the docs to make this cleaner ongoing
* [x] show the SSP version number somewhere in the config settings

https://github.com/simplesamlphp/simplesamlphp/blob/simplesamlphp-1.14/lib/SimpleSAML/Configuration.php#L328

nhoobin:
This is associated with pull request #66.

PR #96, opened 2017-03-01
Issue#25

sam-moodle:
Here is an initial pass at implementing a scheduled task for metadata refresh. For now, I've basically used the same logic that was in the validate_form method of the auth wrapped in some new classes. One major difference is that the metadata is fetched using the Moodle curl class rather than file_get_contents.

I realized as I was creating this PR that I forgot to add a setting to toggle this feature. I can go back and add that, but I'd appreciate a review of what I've done so far. I did have a colleague review it first and made some changes based on his review.

Thanks.


brendanheywood:
@sam-moodle generally looks good just reading it - but there are unit testing failures in some branches can you please sort them?

classes/metadata_fetcher.php
```diff
@@ -0,0 +1,112 @@
...
+            'CURLOPT_NOBODY'         => false,
+        ];
+        $xml = $curl->get($url, $options);
+        $this->curlinfo = $curl->get_info();
+        $this->curlerrorno = $curl->get_errno();
+
+        // If there is a curl errorno from curl_errno().
+        if (!empty($this->curlerrorno)) {
+            $this->curlerror = $xml;
+            throw new \coding_exception('Metadata fetch failed: ' . $xml);
```
brendanheywood: I don't think these should be coding_exceptions, that would imply the developer has done something wrong which isn't true.

classes/metadata_fetcher.php
```diff
@@ -0,0 +1,112 @@
...
+     * @param $url
+     * @return bool
+     * @throws \coding_exception
+     */
+    public function fetch($url, $curl = null) {
+        if (!$curl instanceof \curl) {
+            $curl = new \curl();
+        }
+        $options = [
+            'CURLOPT_SSL_VERIFYPEER' => true,
```
brendanheywood: Have you tested this with self signed or with expired ssl? I've run into issues elsewhere with this in both dev env's and prod. Not sure the best way forward, ideally it would be good to keep this on but explicitly catch this type of failure, especially if cron is going to find the issue later on and it was originally ok. Dunno if it's worth a new config boolean.

classes/metadata_fetcher.php
sam-moodle: Good point. I'll change them to moodle_exceptions.

sam-moodle:
Regarding the unit test failures, Travis is acting up for me so I haven't been able to review them all, but the one I saw was because I'm using Prophecy for mocking objects. IIRC that wasn't available until Moodle 3.0 when PHPUnit was upgraded to 4.x.

classes/task/metadata_refresh.php
```diff
@@ -0,0 +1,147 @@
...
+            mtrace('IDP metadata config not a URL, nothing to refresh.');
+            return;
+        }
+        // Fetch the metadata.
+        if (!$this->fetcher instanceof metadata_fetcher) {
+            $this->fetcher = new metadata_fetcher();
+        }
+
+        try {
+            $rawxml = $this->fetcher->fetch($this->config->idpmetadata);
```
sam-moodle: @brendanheywood, I'm re-thinking these bits. At first, I thought that we wouldn't want the task to fail and reschedule itself, but it may be better to let it go ahead and do that. In that case I wouldn't bother wrapping those calls in try/catch blocks. Instead, just let the exceptions bubble up and have the task scheduler handle.

brendanheywood: Yes I would expect an exception -> fail delay -> monitoring would pick it up - as long as it does not bust the existing good xml.

I would also catch the null case where the saml plugin is not yet configured and the metadata is empty

classes/metadata_fetcher.php
sam-moodle: I have not, though I'm fairly certain that it would fail with self-signed and probably with expired SSL. Seems dangerous to allow someone to switch it off, though I get your point about it being allowed originally. If you really want to be able to switch it off how about a non-UI setting that could be added to config.php to by-pass? That way folks who have been using it and now see an error in their task output can switch it off while they deal with finding out why there is no valid cert.

brendanheywood: I don't really want to switch it off, I just want to make sure the error condition gets reported in a sane way. I've just been caught out by some ssl setups where it works fine in a browser, and even works via curl / wget but then fails in php curl. If it failed I'd want to see a clear exception with 'ssl error' rather than a generic 'fetch error' so people don't chase their tail for an hour.

I like the config.php idea too, sounds good. You almost get it for free with $CFG->forced_plugin_settings

sam-moodle: Made some changes related to this. The way I originally handled cURL errors should propagate the message the the failure was directly related to not being able to verify the peer or the host. I also used $CFG->forced_plugin_settings.

Unrelated to this, how would you like me to handle the unit test failures. They are all related to the fact that I'm using Prophecy to mock objects in the unit tests. That was added to PHPUnit in the version that was added to Moodle 3.0. I'd prefer not to change the use of Prophecy if possible.

sam-moodle: @brendanheywood, any updates on your review?

brendanheywood:
hi @sam-moodle sorry I've been on leave and a little swamped, can you please fix those last travis issues, even if you simply don't run those particular tests on the really old branches and then I'll give it all a once over again

sam-moodle:
No worries @brendanheywood, completely understand. I've updated the PR to skip tests where Prophecy is used when it is not available. Thanks for having another look.

brendanheywood:
@sam-moodle I've just had a chance to test this a bit, running into an issue with this task:

brendanheywood:
oops, take two:

```
brendan@zebrafish 11:00:28 /var/www/moodle.local $ sudo -u www-data php admin/tool/task/cli/schedule_task.php --execute='\auth_saml2\task\metadata_refresh'
Execute scheduled task: Metadata refresh task (auth_saml2\task\metadata_refresh)
... used 6 dbqueries
... used 0.010056972503662 seconds
Scheduled task failed: Metadata refresh task (auth_saml2\task\metadata_refresh),Coding error detected, it must be fixed by a programmer: Setting not added to config class: version
Backtrace:
* line 57 of /auth/saml2/classes/config.php: call to auth_saml2\config->load_config()
* line 70 of /auth/saml2/classes/task/metadata_refresh.php: call to auth_saml2\config->__construct()
* line 146 of /admin/tool/task/cli/schedule_task.php: call to auth_saml2\task\metadata_refresh->execute()
```

The load config seems fairly brittle, it would be nicer if it gracefully just handled any config it wasn't aware of but that might happen to be floating around the db. The particular field it is barfing on is 'version'. Maybe at minimum tone that exception down into a developer debugging call instead?

Also in my config I had left it on the default 'No' to refresh, so I would expect that this task would be completely disabled and do nothing, even if run manually via cli. There is something buggy with the check to honor that config setting, but I think it's related to the exception handling above.

Also can you please add an mtrace call when it successfully updates the xml file too please.

sam-moodle:
Hi @brendanheywood, I'm just back from a couple of days off. I like your suggestion about the developer debugging as that was my intention to begin with: alert the developer that they've added a setting that they may or may not want to be a part of that class.

The task does respect the setting, but loads the config to check it. What you would have seen is the message saying it's disabled.

I'll add the additional mtrace call.

brendanheywood:
hi @sam-moodle,

There is still more issues with the config which need to get sorted, eg now I'm getting this issue:

```
PHP Notice:  Setting not added to config class: alterlogout* line 103 of /auth/saml2/classes/config.php: call to debugging()
```

The root cause it this new object you made to manage config, but you haven't really moved all the code across so now there is two different and competing implementations side by side. Compare this:

https://github.com/sam-moodle/moodle-auth_saml2/blob/issue%2325/auth.php#L42-L56

to this:

https://github.com/sam-moodle/moodle-auth_saml2/blob/issue%2325/classes/config.php#L37-L48

This is redundant, only one of these should exist, so please pick one stick 100% to it. I honestly can't see a heap of value in the new config object, it just looks like a bunch of boilerplate for no real reason. You could have written unit test against an instance of the auth_plugin_saml2 object with no changes at all. If there is a compelling reason to keep it then please convert everything over to use it consistently. But it seems to me to be causing more issues that it's worth and so I'd suggest you ditch it completely.

sam-moodle:
@brendanheywood, thanks for the feedback. I'll remove the class. I initially wanted to use it only for the refresh task, but then thought it might be useful in the auth class as well. I thought making that change might be too much in this PR.

brendanheywood:
Ok this is getting close now, I would still prefer it if there was only a single place that any code touch get_config(). So if you ever want config then you'd do something akin to (new auth_plugin_saml2())->config->blah this way all the keys and defaults will be perfectly consistent.

I'm not sure about these two:

https://github.com/catalyst/moodle-auth_saml2/pull/96/files#diff-845dd473b8219d4b7fd0741d280a3f0fR73
https://github.com/catalyst/moodle-auth_saml2/pull/96/files#diff-845dd473b8219d4b7fd0741d280a3f0fR77

Also there is a bug somewhere in the refresh, if I manually save the config in the gui then it correctly loads then on the login page I get the link:

image

but after running the refresh task this is busted:

image

The root cause is a chunk of duplicated but slightly divergent code here:

https://github.com/sam-moodle/moodle-auth_saml2/blob/issue%2325/classes/task/metadata_refresh.php#L68-L96

vs here:

https://github.com/sam-moodle/moodle-auth_saml2/blob/issue%2325/auth.php#L477-L514

Can you please refactor this so it's all calling the same code path for both gui + task. Technically that code should split across process_config() and validate_form() but I'm ok with it being together in validate to keep it simple

sam-moodle:
@brendanheywood, I fixed the bug and made some minor changes to the task so that it better matches the flow of the validation method, but I did not change the configs. Unfortunately, I'm being asked to change my focus to some other projects. I appreciate the feedback you've given me and the opportunity to work on this plugin. I hope that you're able to move forward with these changes in some way.

brendanheywood:
ok thanks @sam-moodle 

@LoveBug is probably going to pick this up and polish the last bits, so this might get moved to a new pr with code rebased on to yours, we'll soon see

sam-moodle:
Thanks @brendanheywood, let me know if you'd like me to rebase and squash my commits down to one or two commits.

mrkstokking:
Hey Brendan - I think this fell off the radar (ours as well) and we'd like to pick it back up.  Is there any outstanding work required, or is it ready for integration?  We really appreciate your help in getting this issue closed out, thanks!

brendanheywood:
thanks @mrkstokking I think it's fairly close but I'd have to review it again. I'm pretty flat out til after the moot so I'll take a look then.

brendanheywood:
Actually there are some conflicts already, so if you can resolve those and rebase it on top of the 33 branch that would be great, thanks again

sam-moodle:
Sure, I'll take care of rebasing it in the next couple of days.

sam-moodle:
I've rebased against master.

sam-moodle:
Ah, just saw you wanted that rebased against the 33 branch. I'll take of that now.

wjroes:
Hi Sam/Brendan,

I noticed you are loading the plugin config through get_config('auth/saml2') in classes/task/metadata_refresh.php #62 

I don't think that will work any longer (at least in Moodle 3.3+) and it will prevent the task from running.

Willem

sam-moodle:
Thanks @wjroes missed that when I rebased against the 33PLUS branch. I've fixed that and rebased the branch again to fix conflicts w/ version.php

brendanheywood:
I've merged this one, thanks for you patience :)

If you have any time for a port to the 32- branch that would be awesome

Issue #101, opened 2017-04-20
Moodle users that have been suspended or deleted can still login?


praj:
Currently the plugin allows users that have been suspended to login to Moodle (particularly problematic as it doesn't stop them accessing the site after being suspended).

Similarly it allows deleted users to login to Moodle but then throws a PHP catchable fatal error (tested by setting deleted flag to 1) as it can't instantiate the user object properly.

Suggest adding a check for the value of 1 for either the suspended or deleted flags in auth/saml2/auth.php as part of the saml_login() function.

brendanheywood:
hi @praj, thanks - are you able to put together a pull request?

PR #102, opened 2017-04-26
#101 Prevent access to users who are either suspended or deleted

praj:
This is to fix bug #101 which currently allows users to login to Moodle even if their account is either suspended or deleted (though a user with a deleted account will encounter a PHP exception).

### DIFF
--- a/auth.php
+++ b/auth.php
@@ -279,15 +279,15 @@ class auth_plugin_saml2 extends auth_plugin_base {
                 $this->log(__FUNCTION__ . " to lowercase for $key => $uid");
                 $uid = strtolower($uid);
             }
-            if ($user = $DB->get_record('user', array( $this->config->mdlattr => $uid ))) {
+            if ($user = $DB->get_record('user', array( $this->config->mdlattr => $uid, 'deleted' => 0 ))) {
                 continue;
             }
         }
 
-        // Prevent access to users who are either suspended or deleted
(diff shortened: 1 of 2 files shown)


brendanheywood:
Thanks heaps @praj, a few bits of feedback:

* [ ] Conceptually a deleted account is like not having an account in the first place. So I'm thinking that a better approach would be to exclude deleted accounts from the query here:

https://github.com/catalyst/moodle-auth_saml2/pull/102/files#diff-1ce357dc6036d2c85dcc86ba34816421R282

If a user has been deleted, and the autocreate option is on, then I'd expect a new account to be created and for them to get logged in as normal.

* [ ] If the account has been suspended then I think this is worth having a new language pack item saying as much so there is no confusion.

* [ ] can you please make sure travis stays green eg https://travis-ci.org/catalyst/moodle-auth_saml2/builds/225870578

praj:
No worries, I think I'm getting there, let me know how this looks and I think the Travis failures were due to some trailing whitespace which I've removed.

praj:
@brendanheywood are you able to advise why Travis is failing on build jobs 151.6 and 151.15? It seems to be to do with the environment (Postgres) and not necessarily the commit/changes?

brendanheywood:
thanks @praj yeah that postgres error is unrelated, I'll fix that. Code is cool I'll merge it in, thanks!

Issue #103, opened 2017-04-26
saml2 (in Totara) w/ Custom Fields can generate contextid missing errors


dustinbrisebois:
add (line 277): $PAGE (to global list)

add below $auth->requireAuth(); on line 283:

        $context = context_system::instance();
        $PAGE->set_context($context);

this will set the site context for login workflow (in some instances of custom fields with drop down menu items).

brendanheywood:
thanks @dustinbrisebois can you please make a pull request?

roperto:
I am reopening this issue as we had to revert the commit due to issue #117 

Basically this fix was setting the context of a page as system after login with the IDP which made many components in the course page to not work. It probably affected other pages not in the system context as well. After the `require_login()` in `course/view.php` we need to make sure the context is the correct course context.

If you can provide more information on how to replicate this particular issue we can try to find another solution, preferably with Moodle as we cannot automate tests with Totara in Travis.

Thank you,

Daneil

roperto:
I am closing this one as we had several updates to improve its compatibility with Totara and I could not reproduce that one.

Please reopen it if you can still reproduce it.

PR #104, opened 2017-04-27
issue #103 fix context missing from auth flow

dustinbrisebois:


brendanheywood:
@dustinbrisebois can you confirm how to reproduce this bug? Is it on the login page, or on any deep link? And is it only when dual auth is on, or when dual auth is off too?

dustinbrisebois:
@brendanheywood the issue appears to happen in some circumstances when using custom profile fields, the reporter of the issue had a drop down menu custom field that has 4 items in it. It may be due to a collision in the _get_string_ call (which triggered the context error).

brendanheywood:
ok thanks @dustinbrisebois

Issue #132, opened 2017-09-06
Testing issue - setup.php has loose code in its initialiser


roperto:
It should not affect any production code, only tests.

`setup.php` is generally called with `require` but sometimes with `require_once`

If required twice it will throw an exception because of some function defitions are present.

If you need to require it twice (such as in unit tests, where it uses `require_once`) than it was already required in a different test it will not be included, so the test will fail because saml2 was not initialised.

to replicate that problem, in `auth_saml2_locallib_testcase` add the following test:

```php
    public function test_it_makes_next_fail() {
        global $CFG;
        $this->resetAfterTest(true);
        self::setAdminUser();
        require_once($CFG->libdir.'/adminlib.php');
        admin_get_root();
    }
```

`admin_get_root()` will initialise the admin tree, which will end up initialising SAML2 (it could be a test in core or anywhere else that used that function). Now when SAML2 test `test_auth_saml2_sp_metadata` runs it will fail.

roperto:
Fixed by creating setuplib.php

PR #135, opened 2017-09-07
Issue #132: Moved setup.php definitions to setuplib.php so setup.php …

roperto:
…can be required more than once.


brendanheywood:
+1

PR #143, opened 2017-09-29
Added support for using Redis as the store

sam-moodle:
This PR adds support for using Redis as the store type. It uses 2 new $CFG variables that can be set in Moodle's config.php: $CFG->auth_saml2_store and $CFG->auth_saml2_redis_server.

To use it you would add something like this to the config file:

$CFG->auth_saml2_store = '\auth_saml2\redis_store';
$CFG->auth_saml2_redis_server = '127.0.0.1';

To run PHPUnit tests, AUTH_SAML2_REDIS_STORE_TEST_SERVER must be defined in the Moodle config file, set to a Redis connection string.

### DIFF
--- a/classes/redis_store.php
+++ b/classes/redis_store.php
@@ -1,0 +1,149 @@
+<?php
+// This file is part of Moodle - http://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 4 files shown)


brendanheywood:
Looks great @sam-moodle, sorry for the delay post moot I'll get to your other pr's soonish too

PR #145, opened 2017-10-09
auth_saml2: Allow the use of a discovery service URL

sam-moodle:
This allows the use of a discovery service via a URL. As the changes are presented in this pull request the feature is enabled by setting $CFG->auth_saml2_disco_url to a string that is a URL for the discovery service being used.

When the above config is set the hosted SP discoURL config is set the value it holds and the hosted SP config idp is set to null (see https://simplesamlphp.org/docs/stable/saml:sp). When idp is null and the authentication process is started a discovery service operation is started.

We have plans to expose this via the auth_saml2 settings page in the future, but when we needed to implement this we were constrained to making non-UI changes only, hence the config.php setting.

### DIFF
--- a/config/authsources.php
+++ b/config/authsources.php
@@ -36,7 +36,8 @@ $config = array(
     $saml2auth->spname => array(
         'saml:SP',
         'entityID' => "$wwwroot/auth/saml2/sp/metadata.php",
-        'idp' => $saml2auth->config->entityid,
+        'discoURL' => !empty($CFG->auth_saml2_disco_url) ? $CFG->auth_saml2_disco_url : null,
+        'idp' => empty($CFG->auth_saml2_disco_url) ? $saml2auth->config->entityid : null,
         'NameIDPolicy' => null,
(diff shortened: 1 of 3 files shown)


brendanheywood:
@sam-moodle thanks again :)

Did you get anywhere with a gui for this one? Also if you can also back port to the 3.2- stable branch that would be awesome. There is a growing list of config.php only features so I've started a new section in the readme for these:

https://github.com/catalyst/moodle-auth_saml2#configuration

If you add more can you please touch the readme in future pr's

sam-moodle:
@brendanheywood you're welcome. We haven't added the UI for it yet, but it is still on our radar and we'll create a PR here when we have it ready.

I'll work on the back port in the next couple of days. If we add any other config.php only features I'll be sure to update the readme.

sam-moodle:
@brendanheywood here is that backport: https://github.com/catalyst/moodle-auth_saml2/pull/163

PR #156, opened 2017-11-03
Add ability to configure SSPHP to log to file

catalystfd:
The debug messages from SSPHP can be very verbose, so having this as an
option is very useful for extensive debugging sessions


catalystfd:
rebased

PR #158, opened 2017-11-03
Config setting to enable exposing NameID as attribute (Moodle 3.3)

catalystfd:
Often for SSO the IDP will expose a NameID field only, so this
configuration setting will expose the NameID field as an attribute
'nameid' which can then be matched against for login

### DIFF
--- a/settings.php
+++ b/settings.php
@@ -88,6 +88,12 @@ if ($ADMIN->fulltree) {
             get_string('logdirdefault', 'auth_saml2'),
             PARAM_TEXT));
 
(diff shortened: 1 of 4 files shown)


settings.php
```diff
@@ -75,6 +75,13 @@
             get_string('debug_help', 'auth_saml2', $CFG->wwwroot . '/auth/saml2/debug.php'),
             0, $yesno));
 
+    //Add NameID as attribute
```
optimumtact: need to add a space here

catalystfd:
rebased

Issue #161, opened 2017-11-23
Honor moodle proxy settings


brendanheywood:
**What happened?**

Moodle has optional proxy settings found here:

/admin/settings.php?section=http

**What you expected:**
* [ ] when making the idp xml curl call, honor those settings

brendanheywood:
Note done in 32- via:

https://github.com/catalyst/moodle-auth_saml2/pull/160

Not yet done in 3.3, needs to be done differently inside:

https://github.com/catalyst/moodle-auth_saml2/blob/MOODLE_33PLUS/classes/metadata_fetcher.php#L59

PR #164, opened 2017-11-30
Backport issue 25 IDP metadata refresh

sam-moodle:


sam-moodle:
@brendanheywood I had to do some rebasing/conflict resolution, etc for this and haven't had a chance to pull this branch into a site to test. Just FYI.

brendanheywood:
@sam-moodle one thing to resolve, really in both stables is honoring the moodle proxy settings. Because the metadata fetch does it's own curl it misses out on this. Recently in parallel support for this landed on the 32- branch so this PR would be a regression:

https://github.com/catalyst/moodle-auth_saml2/pull/160/commits/e613c0f06ce59227fde38db44fcab007cfb1bc79

So we could add more code to manually honor the proxy, or we could instead use the moodle curl object, or just simplify it and just use download_file_content()

See also this issue:

https://github.com/catalyst/moodle-auth_saml2/issues/161

Thoughts?

sam-moodle:
@brendanheywood I'll double check, but I think the metadata_fetcher class does use the Moodle curl class so the proxy settings should be respected.

version.php
```diff
@@ -24,8 +24,8 @@
 
 defined('MOODLE_INTERNAL') || die();
 
-$plugin->version   = 2017081700;    // The current plugin version (Date: YYYYMMDDXX).
-$plugin->release   = 2017081700;    // Match release exactly to version.
+$plugin->version   = 2017101700;    // The current plugin version (Date: YYYYMMDDXX).
```
sam-moodle: @brendanheywood I'm not sure what your version bumping strategy is, but I just realized that this version is higher than the version in the 3.3 branch that updates the configs to the 3.3 style.

sam-moodle:
@brendanheywood I did double check the metadata_fetcher class. It uses Moodle's curl class. Let me know about the version bump and I can fix that.

version.php
brendanheywood: Ah crap yeah, I think there is a larger problem with the two stables using todays date instead of the fork date + increment. ala https://moodle.org/mod/forum/discuss.php?d=275635#p1184986

I think we'll just have to pick two arbitrary dates for each stable instead of the fork date and then increment from there.

PR #189, opened 2018-02-13
Fixes #129 Login overwrites description field.

### LINKED ISSUE #129: Login overwrites description field with 1
When doing a login, if a user has any fields configured to update on every login, the description gets overwritten with a 1.

This occurs at line 318: https://github.com/catalyst/moodle-auth_saml2/blob/MOODLE_33PLUS/auth.php#L318 on the call to get_complete_user_data.  Moodle defines that method as being used for getting session info and not for truly pulling user data.  Please see the following for exact line this occurs:
https://github.com/moodle/moodle/blob/8146b1f06de51d1ca6b4043bc524ed2fe26aaa72/lib/moodlelib.php#L4717

For a temporary fix, I have commented out lines 318 and 393 as profile_save_data requires the full user data that is not fetched by removing 318.  This works for this site as we have no custom fields being populated, but is not ideal.

I wanted to log this here so anyone else having the issue can see it.  I will work some more to find a full fix, unless someone already has an idea!

Thanks!

danmarsden:
Ran into this today - can someone review this to check it makes sense?


roperto:
Hi @danmarsden , sorry for taking some time to get back to you, we had some major improvements happening in this plugin and we merged all branches into one.

I am reopening this into the correct branch, we should merge it very soon.

roperto:
Closing this in favour of #193

PR #217, opened 2018-03-08
Closes #213 - Implements missing CI tests, fixes all errors and some warnings.

roperto:

PR #219, opened 2018-03-09
Inlining constant TABLE_KVSTORE into code, see #217

roperto:
FYI dan , merging directly when green :-)


danmarsden:
cool - thanks!

PR #250, opened 2018-04-19
Fixes #249 - nosinglelogout

### LINKED ISSUE #249: Do not attempt to log single log out

roperto:


roperto:
I am closing the PR because during an upgrade it is automatically making it default to not sign out, which can potentially be a security issue having this behaviour changed.

We need to fix it before merging.

PR #251, opened 2018-04-19
Add support for Noredirect param.

danmarsden:
The noredirect param has been used historically by other Moodle authentication plugins to bypass a redirected login page. "Magicpony" as a parameter for no redirect support was also around (i think by Mjollnir)  - but we probably don't need to add support for that... :-)


roperto:
Hi @danmarsden -- It looks good to be and we have CI green for your PR, can it be merged?

danmarsden:
cool thanks for the review - merged.

PR #258, opened 2018-05-31
add $PAGE, login context for some profile fields, and set_url for redirect

dustinbrisebois:
- some profile field instances require context to be set before field call -- this is fixed removing the $PAGE->set_context required error
- on redirect the system now requires $PAGE->set_url to be called prior to redirect - this is fixed.

### DIFF
--- a/version.php
+++ b/version.php
@@ -24,7 +24,7 @@
 
 defined('MOODLE_INTERNAL') || die();
 
-$plugin->version   = 2018032900;    // The current plugin version (Date: YYYYMMDDXX).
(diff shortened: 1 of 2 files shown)


version.php
```diff
@@ -24,7 +24,7 @@
 
 defined('MOODLE_INTERNAL') || die();
 
-$plugin->version   = 2018032900;    // The current plugin version (Date: YYYYMMDDXX).
+$plugin->version   = 20180531000;    // The current plugin version (Date: YYYYMMDDXX).
```
roperto: Ops this one has one extra zero!

PR #262, opened 2018-06-12
Do not fail if driver does not support screenshot.

roperto:


roperto:
If we get a failed step, we get the message:

```
--- Failed hooks:

    AfterStep # behat_auth_saml2::take_screenshot_after_failure()
      Screenshots are not supported by Behat\Mink\Driver\GoutteDriver (Behat\Mink\Exception\UnsupportedDriverActionException)
```

fillowed by the failed step. It is not a problem per se because it will only "fail" to take a screenshot if a step is failing.

However, some people may point the error to SAML plugin instead of addressing the original error, so if there is problem taking a screenshot just log the message and ignore it

PR #266, opened 2018-06-20
Adding multiple IdP support for a single metadata entry

rhell4:


config/authsources.php
```diff
@@ -38,8 +38,29 @@
...
+}
+
 if (!empty($SESSION->saml2idp)) {
     foreach ($saml2auth->idpentityids as $idpentityid) {
+        if (!is_string($idpentityid)) {
+            $subidps = (array)$idpentityid;
+            foreach ($subidps as $subidp => $active) {
+                if ($SESSION->saml2idp === md5($subidp)) {
+                    $idp = $subidp;
+                    break 2;
```
roperto: Minor thing: when we need to break an outer loop like that there is probably a way to simplify it.

tests/phpunit/metadata_refresh_test.php
```diff
@@ -68,123 +69,7 @@ public function test_metadata_refresh_idpmetadata_notconfigured() {
...
         $refreshtask = new metadata_refresh();
 
+        $this->expectOutputString('IdP metadata not configured.' . "\n");
         self::assertFalse($refreshtask->execute());
     }
-
-    /**
-     * @expectedException \moodle_exception
-     */
-    public function test_metadata_refresh_fetch_fails() {
```
roperto: Do we need to remove those tests? Not sure why.

PR #273, opened 2018-07-11
Adding a setting for signing algorithm, defaulting it to sha256

adamlynam-catalyst:
Essentially a Moodlfied version of:

https://github.com/MaharaProject/mahara/commit/059b0765e4ce6867cccfda3c6b1d5426fe2c878d#diff-0e96462bae3baa9606309828e2638802

I should make it clear, this exposes one setting that controls both the signing algorithm for the service provider and the digest algorithm for generating self-signed certificates for the service provider.

### DIFF
--- a/classes/ssl_algorithms.php
+++ b/classes/ssl_algorithms.php
@@ -1,0 +1,68 @@
+<?php
+// This file is part of Moodle - http://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+/**
+  * @package    auth_saml2
+  * @author     Adam Lynam <adam.lynam@catalyst.net.nz>
+  * @copyright  Catalyst IT
+  * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+  */
+
+namespace auth_saml2;
+
+defined('MOODLE_INTERNAL') || die();
+
+global $saml2auth;
+
+abstract class ssl_algorithms {
+  /*
+   * Return a sensible default signature algorithm for simplesamlphp config
+   */
+  public static function get_default_saml_signature_algorithm() {
+      //Sha1 is deprecated so we default to something more sensible
+      return 'http://www.w3.org/2001/04/xmldsig-more#rsa-sha256';
+  }
+
+  /*
+   * Return an array of signature algorithms in a form suitable for feeding into a dropdown form
+   */
+  public static function get_valid_saml_signature_algorithms() {
+      $return = array();
+      $return['http://www.w3.org/2001/04/xmldsig-more#rsa-sha256'] = get_string('sha256', 'auth_saml2');
+      $return['http://www.w3.org/2001/04/xmldsig-more#rsa-sha384'] = get_string('sha384', 'auth_saml2');
+      $return['http://www.w3.org/2001/04/xmldsig-more#rsa-sha512'] = get_string('sha512', 'auth_saml2');
+      $return['http://www.w3.org/2000/09/xmldsig#rsa-sha1'] = get_string('sha1', 'auth_saml2');
+      return $return;
+  }
+
+  /*
+   * Return an array of digest algorithms in a form suitable for feeding into a dropdown form
+   */
+  public static function convert_signature_algorithm_to_digest_alg_format($signature_algorithm) {
+      switch($signature_algorithm) {
+          case 'http://www.w3.org/2001/04/xmldsig-more#rsa-sha256':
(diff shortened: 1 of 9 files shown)


adamlynam-catalyst:
Yay it works.

setuplib.php
```diff
@@ -41,7 +43,13 @@
 function create_certificates($saml2auth, $dn = false, $numberofdays = 3650) {
     global $SITE;
 
-    $opensslargs = array();
+    $signaturealgorithm = ssl_algorithms::get_default_saml_signature_algorithm();
+    if (isset($saml2auth->config->signaturealgorithm)) {
```
dmitriim: Should we check if it's not empty here?

classes/ssl_algorithms.php
```diff
@@ -0,0 +1,68 @@
...
+                return 'SHA256';
+            case 'http://www.w3.org/2001/04/xmldsig-more#rsa-sha384':
+                return 'SHA384';
+            case 'http://www.w3.org/2001/04/xmldsig-more#rsa-sha512':
+                return 'SHA512';
+            case 'http://www.w3.org/2000/09/xmldsig#rsa-sha1':
+                return 'SHA1';
+        }
+
+        return 'SHA256';
```
dmitriim: Are we sure SHA256 is a deafult all the time?

setuplib.php
adamlynam-catalyst: I struggled with what checks needed to be here.

In the tests it was not being set, but I wasn't sure how it got into that state. Is there actually a way for a config variable to be empty string or not set.

dmitriim: Moodle is unpredictable :) But looking at the rest of the code this line should be good.

config/config.php
```diff
@@ -82,6 +84,8 @@
 
     'enable.http_post' => false,
 
+    'signature.algorithm' => $saml2auth->config->signaturealgorithm,
```
dmitriim: Should we check if 'signaturealgorithm' set and not empty here? and also get a default one if it's not set or empty for some reason?

lang/en/auth_saml2.php
```diff
@@ -140,3 +140,13 @@
  * Privacy provider (GDPR)
  */
 $string["privacy:no_data_reason"] = "The Saml2 authentication plugin does not store any personal data.";
+
+/*
+ * Signing Algorithm
+ */
+$string['sha1'] = 'Legacy SHA1 (Dangerous)';
```
dmitriim: "(Dangerous)" love it :) :+1:

classes/ssl_algorithms.php
```diff
@@ -0,0 +1,68 @@
...
+ * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+
+namespace auth_saml2;
+
+defined('MOODLE_INTERNAL') || die();
+
+global $saml2auth;
+
+abstract class ssl_algorithms {
```
dmitriim: A simple tests fro this class possible could help in the future...

classes/ssl_algorithms.php
adamlynam-catalyst: Are you thinking that maybe I should fall back to `get_default_saml_signature_algorithm()` before a hard coded default?

I was thinking it would be unlikely there wasn't a mapping that applied in the `switch`.

```
public static function convert_signature_algorithm_to_digest_alg_format($signaturealgorithm) {
    switch($signaturealgorithm) {
        case 'http://www.w3.org/2001/04/xmldsig-more#rsa-sha256':
            return 'SHA256';
        case 'http://www.w3.org/2001/04/xmldsig-more#rsa-sha384':
            return 'SHA384';
        case 'http://www.w3.org/2001/04/xmldsig-more#rsa-sha512':
            return 'SHA512';
        case 'http://www.w3.org/2000/09/xmldsig#rsa-sha1':
            return 'SHA1';
    }

    switch(self::get_default_saml_signature_algorithm()) {
        case 'http://www.w3.org/2001/04/xmldsig-more#rsa-sha256':
            return 'SHA256';
        case 'http://www.w3.org/2001/04/xmldsig-more#rsa-sha384':
            return 'SHA384';
        case 'http://www.w3.org/2001/04/xmldsig-more#rsa-sha512':
            return 'SHA512';
        case 'http://www.w3.org/2000/09/xmldsig#rsa-sha1':
            return 'SHA1';
    }

    return 'SHA256';
}
```

dmitriim: I think it's all good. Just wasn't sure if SHA256 should be a default one in all cases. In theory(not likely) in the future we can add more lines to get_valid_saml_signature_algorithms() and forget to update convert_signature_algorithm_to_digest_alg_format(). In this case you will get broken mapping/converting which could cause some pain.

config/config.php
adamlynam-catalyst: Sounds like a sensible thing to do. I will add that.

lang/en/auth_saml2.php
adamlynam-catalyst: I would love to take credit for that, but it was @catalystfd who put that into the Mahara version:

https://github.com/MaharaProject/mahara/commit/059b0765e4ce6867cccfda3c6b1d5426fe2c878d#diff-22db7a49684162f53954690a821df437R83

classes/ssl_algorithms.php
adamlynam-catalyst: I would love to add some tests. I tried to follow the convention for phpunit tests in `tests/phpunit`, but it seems very different from the phpunit I have worked with before.

I cannot work out how to actually run the tests. Is there a way to just use the `phpunit` CLI?

PR #287, opened 2018-09-26
add login/logout page callbacks

tuanngocnguyen:


auth.php
```diff
@@ -313,6 +313,14 @@ public function pre_loginpage_hook() {
      * All the checking happens before the login page in this hook
      */
     public function loginpage_hook() {
+
+        $pluginsfunction = get_plugins_with_function('auth_saml2_loginpage_hook', 'lib.php');
```
dmitriim: we claim support of 2.7 moodle for this plugin (https://moodle.org/plugins/pluginversions.php?plugin=auth_saml2).
Does get_plugins_with_function exists in Moodle 2.7 ?

auth.php
```diff
@@ -755,5 +740,20 @@ public function get_idp_cookie() {
             return $_COOKIE[$cookiename];
         }
     }
+
+    /**
+     * @param $function function name
```
dmitriim: Would be good to see some documentation fro the function and its params. Thanks

PR #302, opened 2018-12-13
idp empty string fix & $PAGE context fixes

dustinbrisebois:
there are two problems wrapped up in here one is that $PAGE context can have strange results IF the PAGE requested requires context settings

Second, is that if the IDP line is empty then the system tracks it as a valid IDP value for processing. There are two tracks here in idp_parser, one is if the line is EMPTY the second is if the idpurl is exactly http (as it should be a FQDN not just an empty address).

### DIFF
--- a/auth.php
+++ b/auth.php
@@ -428,7 +428,7 @@ class auth_plugin_saml2 extends auth_plugin_base {
     public function saml_login() {
 
         // @codingStandardsIgnoreStart
-        global $CFG, $DB, $USER, $SESSION, $saml2auth;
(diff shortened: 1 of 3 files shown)


brendanheywood:
@dustinbrisebois can you please outline the issue in a little more detail that this fixes?

katerynadegtyariova:
There are a few problems with this PR:

1. No clear description of the problem the patch is supposed to fix and therefore no way of reproducing the problem.
2. Merge conflicts.
3. No test case to test that the patch is working.

For this reason this PR is closed.
Please feel free to reopen it if the above problems are resolved.

PR #313, opened 2019-02-12
Flag accounts

tomdickman:
This feature adds a new section of admin settings to the Moodle SAML2 plugin to allow administrators to set a flag attribute to look for when authenticating a user from an IdP, and an action to be taken if the flag is present and of a particular value.

The requirement was to intercept an SSO Identity Provider (IdP) login process for Moodle as a component of the SimpleSAMLphp Service Provider (SP), which is a component of this SAML2 plugin, and conduct the following:

* check if a flag attribute has been set
* validate if the value of this attribute in the IdP authentication indicates that the user is flagged
* take action to either
  
  * display a custom message
  * redirect the user to an external URL

Tests have been included for URL validation of the external URL and behaviour testing for the new menu settings.


classes/settings_helper.php
```diff
@@ -0,0 +1,75 @@
...
+     * Direct initiation not allowed, utility class.
+     */
+    protected function __construct() {
+    }
+
+    /**
+     * Setter for $http_https_regex.
+     */
+    protected static function set_https_regex() {
+        // Build the regular expression for validating https/https URLs in accordance with RFC3986.
```
mattporritt: could you use a core PHP function over a regex? Will http://php.net/manual/en/function.filter-var.php work?

tomdickman: The URL is handled by the adminlib, through a call to `$settings->add` as a parameter. I tried using the `PARAM_URL` built in constant and the `PARAM_HOST` constant but neither worked to validate the scheme as http/https in testing (despite `PARAM_HOST` docs indicating it should do exactly this). So the decision was made to use regex and validate through testing of a large range of valid URLs.

classes/settings_helper.php
```diff
@@ -0,0 +1,75 @@
...
+    /**
+     * @var string representation of PCRE Regular Expression for checking http/https scheme URLs
+     * in accordance with RFC3986.
+     */
+    private static $httpsregex;
+
+    /**
+     * Direct initiation not allowed, utility class.
+     */
+    protected function __construct() {
```
dmitriim: you can get rid of empty constructor

auth.php
```diff
@@ -545,6 +553,51 @@ public function saml_login() {
...
+    /**
+     * Checks if the flagged user feature is enabled and if so, handles configured flag attribute
+     *
+     * @param $attributes
+     *
+     * @throws \moodle_exception
+     */
+    protected function handle_flagged_login($attributes) {
+
+        if (!isset($attributes) || empty($this->config->flagattribute)) {
```
dmitriim: !isset($attributes) doesn't make any sense. PHP will die as your function needs attributes as a parameter

auth.php
```diff
@@ -545,6 +553,51 @@ public function saml_login() {
...
     }
 
+    /**
+     * Checks if the flagged user feature is enabled and if so, handles configured flag attribute
+     *
+     * @param $attributes
+     *
+     * @throws \moodle_exception
+     */
+    protected function handle_flagged_login($attributes) {
```
dmitriim: Don't you think this function needs a bit of love? 
Even with existing comments it's really hard to follow what's going on there. 
Basically we need:
1. Check if  flagattribute set at all in configs. If not - > log, terminate handle. 
2. Check if flagattribute is exist in $attributes at all.  If not - > terminate handle.
3. Get flagattribute from $attributes and check if it's the same as configured value.
4, Then handle redirect based on isflagactive and configured flagresponsetype.

"1" and "2" could be abstracted to a validation function with a meaningful name.
"3" could be abstracted to  a function with a meaningful name.

This I recon will make the code much cleaner.

auth.php
tomdickman: Removed

classes/settings_helper.php
tomdickman: Going to leave as is, best solution due to the restrictions of core Moodle admin settings and the work involved with developing a workaround to extend the class `admin_setting_configselect`

auth.php
tomdickman: Refactored into smaller methods with individual concerns, could you please review changes @dmitriim ?

tomdickman:
This has been close and the branch has been rebased and moved to a new pull request.

PR #315, opened 2019-02-14
Multiple idp improvements

rhell4:
This includes a bunch of improvements for multiple IdPs such as:

- Option for creating aliases for IdPs so that they can be directly used by providing a url param
- In addition to the above, a default option which defaults to an IdP unless the above alias is provided
- Option for automatically creating admin accounts on login with a certain IdP
- Moving the multiple IdP structure from json arrays to actual database entries


config/authsources.php
```diff
@@ -37,36 +37,19 @@
...
-        if (!is_string($idpentityid)) {
-            $subidps = (array)$idpentityid;
-            foreach ($subidps as $subidp => $active) {
-                if ($SESSION->saml2idp === md5($subidp)) {
-                    $idp = $subidp;
-                    break 2;
-                }
+    foreach ($saml2auth->metadataentities as $idpentities) {
+        foreach ($idpentities as $md5entityid => $idpentity) {
+            if ($SESSION->saml2idp === $md5entityid) {
```
brendanheywood: it feels wrong to be using the session object here in the config?

brendanheywood:
@rhell4 this barfed on the upgrade:

Exception - Call to a member function add_field() on null

More information about this error

×Debug info: 
×Stack trace:
line 224 of /auth/saml2/db/upgrade.php: Error thrown
line 636 of /lib/upgradelib.php: call to xmldb_auth_saml2_upgrade()
line 1852 of /lib/upgradelib.php: call to upgrade_plugins()
line 694 of /admin/index.php: call to upgrade_noncore()

brendanheywood:
@rhell4 can you please fix the build matrix for https://travis-ci.org/catalyst/moodle-auth_saml2/jobs/498506351

brendanheywood:
@rhell4 still not green?

rhell4:
@brendanheywood looks like it's all green now

Issue #320, opened 2019-04-03
Unit test failure in totara 10 test_all_install_xml_files_formatted_correctly


kristian-94:
**What happened?**
Installed master branch into a clean totara 10.19

Run test 
`vendor/bin/phpunit tool_xmldb_instal_xml_testcase admin/tool/xmldb/tests/install_xml_test.php`
We see 

>  There was 1 failure:
> 
> 1) tool_xmldb_instal_xml_testcase::test_all_install_xml_files_formatted_correctly
> XMLDB file '/siteroot/auth/saml2/db/install.xml' is different to the generated version, please edit the file in XMLDB editor and save it.
> Failed asserting that false is true.
> 
> /siteroot/admin/tool/xmldb/tests/install_xml_test.php:52
> /siteroot/lib/phpunit/classes/base_testcase.php:600
> /siteroot/lib/phpunit/classes/advanced_testcase.php:68
> 
> To re-run:
>  vendor/bin/phpunit tool_xmldb_instal_xml_testcase admin/tool/xmldb/tests/install_xml_test.php
> 
> FAILURES!
> Tests: 1, Assertions: 71, Failures: 1.
>

PR #321, opened 2019-04-03
Change install.xml and file perms to fix unit tests. Fixes issue #320 and #323

kristian-94:
Minor changes to install.xml that fix unit tests in totara 10.19, see issue #320


danmarsden:
I think I would have preferred to see this was done using xmldb editor rather than hand-editing the install.xml... any chance you could open this in moodle's xmldb editor and use that to save it to make sure it has fixed anything else up?

kristian-94:
> I think I would have preferred to see this was done using xmldb editor rather than hand-editing the install.xml... any chance you could open this in moodle's xmldb editor and use that to save it to make sure it has fixed anything else up?

Hi @danmarsden , I did use the xmldb editor to make this change, the only thing I removed was the version date that it changed, `<XMLDB PATH="auth/saml2/db" VERSION="20180308"`
I wanted to change as little code as possible to get it passing this unit test.

kristian-94:
@danmarsden I noticed that travis is failing because of the new permissions that the unit test requires on the files, the travis error is: 

> $ phpenv config-rm xdebug.ini
> File xdebug.ini removed from /home/travis/.phpenv/versions/7.0/etc/conf.d.
> before_install.5
> 0.02s$ echo "extension = redis.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini
> 0.00s$ tests/idp/simplesamlphp.sh
> /home/travis/.travis/functions: line 104: tests/idp/simplesamlphp.sh: Permission denied
> The command "tests/idp/simplesamlphp.sh" failed and exited with 126 during .

I wonder what you think is the best way to resolve that conflict?

danmarsden:
Haven't seen that before sorry - bit crazy busy so haven't had a chance to investigate - the patch makes sense though, and so I've merged it anyway. In future I'd suggest leaving the version date change because it shows that you made the change using xmldb to the person reviewing the patch :-)

PR #326, opened 2019-04-22
Group attribute. Allow/deny login is decided on user membership.

dlnsk:


auth.php
```diff
@@ -597,6 +602,30 @@ public function saml_login() {
...
     }
 
+    /**
+     * Testing user's groups attribute and allow access decided on preferences.
+     *
+     */
+    public function is_access_allowed_for_member($attributes) {
+
+        if (empty($this->config->groupattr)) {
+            $this->log(__FUNCTION__ . " group attribute not found.");
```
danmarsden: should there really be a logging event here? - or was this left over from initial debuging?

auth.php
```diff
@@ -597,6 +602,30 @@ public function saml_login() {
...
+        }
+
+        $groups = $attributes[$this->config->groupattr];
+        $deny  = preg_split("/[\s,]+/", $this->config->restricted_groups, null, PREG_SPLIT_NO_EMPTY);
+        $allow = preg_split("/[\s,]+/", $this->config->allowed_groups, null, PREG_SPLIT_NO_EMPTY);
+        if (!empty(array_intersect($deny, $groups)) || empty(array_intersect($allow, $groups))) {
+            $this->log(__FUNCTION__ . " user '$uid' is in restricted group or isn't in allowed. Access denied.");
+            return false;
+        } else {
+            $this->log(__FUNCTION__ . " user '$uid' is in allowed group and isn't in restricted. Access allowed.");
```
danmarsden: also - unsure if we really need to be logging this here?

dlnsk: I have written in style of the module. Please, see around. Here are many logging.
I'm also thing that when somebody can't login the simplest way is to see in log what's wrong.
I can remove any logging if you want but my opinion is this is wrong way.

danmarsden: seems a bit inconsistent when looking through the code - IMO it seems a bit excessive to log that $this->config->groupattr is empty on every call - it's going to fill up the log quite quickly when a site doesn't use that param. I'll leave it for @brendanheywood to make a call on :-)

dlnsk: Agree with groupattr

danmarsden:
cool - looks good to me - merging. travis is failing on master but the PHP 7.2 pull request which I still haven't tested on a prod server will fix that - hopefully we'll get to that this week.

PR #328, opened 2019-05-07
Feature flag accounts

tomdickman:
This feature adds a new section of admin settings to the Moodle SAML2 plugin to allow administrators to set a flag attribute to look for when authenticating a user from an IdP, and an action to be taken if the flag is present and of a particular value.

The requirement was to intercept an SSO Identity Provider (IdP) login process for Moodle as a component of the SimpleSAMLphp Service Provider (SP), which is a component of this SAML2 plugin, and conduct the following:

- check if a flag attribute has been set
- validate if the value of this attribute in the IdP authentication indicates that the user is flagged
- take action to either:
  - display a custom message
  - redirect the user to an external URL

Tests have been included for URL validation of the external URL and behaviour testing for the new menu settings.


brendanheywood:
Note: this should NOT be merged - just reiterating my previous feedback that I would prefer not to see this land in saml2, it should be orthogonal to the auth. ie saml2 simply passes on some attributes and / or sets user roles, and then this second plugin does something special with those attributes or role

dmitriim:
Will be worked out as part of https://github.com/catalyst/moodle-auth_saml2/pull/360

PR #332, opened 2019-06-12
fix auth saml2 idp test functions for multiple idps

mattporritt:


test.php
```diff
@@ -88,7 +88,6 @@
     var_dump($attributes);
     echo 'IdP: ' . $auth->getAuthData('saml:sp:IdP');
     echo '</pre>';
-    echo '<p>You are logged in: <a href="?logout=true">Logout</a></p>';
```
brendanheywood: Was this intentional? We need to be able to test logout in isolation the same as login

mattporritt: Sort of, I probably shouldn't have committed it.  On my test rig clicking that link through an error, I removed it when I was testing.  Probably need to get to the bottom of why it errors

mattporritt:
Hi @brendanheywood,
That test script was broken in some special ways.  However, test IDP logout now also works.

Cheers,
Matt P

PR #341, opened 2019-07-29
setting for logdir changed to configexecutable 

james-voong:
so that $CFG->preventexecpath can lock it


brendanheywood:
thanks @james-voong  I'm going to close this as a 'won't do' and this setting isn't an executable its a path, so $CFG->preventexecpath shouldn't apply to it. If you want to lock it you've always been able to do that via config.php:

$CFG->forced_plugin_settings['auth_saml2']['logdir'] = 'path';

Issue #338, opened 2019-07-24
Inconsistent idp value


davosmith:
https://github.com/catalyst/moodle-auth_saml2/blob/04044eabec6bf0841d8f11f5a0da4abed53ae883/config/authsources.php#L43

When doing an unsolicited SSO direct from the IDP, with browser data cleared, it fails the first time, then succeeds the second time.

As far as I can tell, the reason it fails the first time is that when $SESSION->saml2idp is _not_ set, the 'idp' value in the config returned from the file I've indicated above is the **md5 hash** of the 'entityid' value. When $SESSION->saml2idp _is_ set, then the 'idp' value in the config is the raw 'entityid' value (not the md5 hash of that value). Please compare line 43, linked above, with line 49 in the same file.

This means that when the config is later retrieved and the entityid value is checked against the idp value, it fails to match and throws an error.

As far as I can tell, the correct fix would be to simply remove the md5() call from line 43.

samchaffee:
@davosmith, we're experiencing the same issue with unsolicited SSO direct from the IdP. Have you  been running the suggested patch since this was reported?

davosmith:
@samchaffee we've been running this patch on a UAT site, for testing purposes, but not yet deployed to a production site.

samchaffee:
Thanks, @davosmith, I appreciate it. We'll likely be putting this in place as well and I'll update this thread with any feedback we have if it's helpful for other folks.

kristian-94:
Thank you @davosmith for your comment here that helps to track down the issue. I'm able to reproduce the issue as you described, by simply clearing the browsers data and then trying to log in with saml. I'll be working on a patch to fix this issue.

dmitriim:
Hi all, 
Have you tried to resave config of the plugin? Does it fix the issue?
It seems like it's related to https://github.com/catalyst/moodle-auth_saml2/issues/362

davosmith:
Hi @dmitriim I'm not convinced that is the same issue - this isn't to do with the loading of saved config from the database or caches, but the settings that are saved temporarily in the $SESSION global. There is simply a mismatch between the idp identifier between two different places in the code where the relevant details in the $SESSION are initialised.

The fix I've already proposed is deployed and working for our customer on their UAT and production sites, so I'm not planning on doing any further investigation at the moment.

brendanheywood:
This should be fixed now

PR #343, opened 2019-08-08
Do not hash idp entity data when session data is not set

kristian-94:
- Issue #338 
We didn't hash this retrieved idp entity data before the multiple idp changes: https://github.com/catalyst/moodle-auth_saml2/blob/0defa84f3ba212577e512700f696ecc0a9ec2212/config/authsources.php#L41

But now we are hashing with md5, which causes issues. To see the problem with the master branch of the plugin:

1. Refresh the saml login page -- adds idp entity to session data
2. Clear browser data -- remove this session data
3. Login with saml auth credentials

You will see the following error: 

> SAML2 exception: Cannot retrieve metadata for IdP 'http://...' because it isn't a valid IdP for this SP.

This change fixes the above workflow and prevents this error.

### DIFF
--- a/config/authsources.php
+++ b/config/authsources.php
@@ -40,7 +40,7 @@ $config = [];
 $arr = array_reverse($saml2auth->metadataentities);
 $metadataentities = array_pop($arr);
 $idpentity = array_pop($metadataentities);
-$idp = md5($idpentity->entityid);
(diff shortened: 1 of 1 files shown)


kristian-94:
On further inspection, the error still can occur under certain conditions even with this fix. It will require more investigation.

PR #344, opened 2019-08-08
Cannot retrieve metadata for IdP Issue #338

kristian-94:
For issue #338 
### An update
This has been tested and is still an issue in the latest master branch. 
To see the problem:
visit your site's saml login page
clear browser 'Cookies and other site data'
go back to your site's saml login page and log in
you will see the following error: 

> SAML2 exception: Cannot retrieve metadata for IdP 'http://adfs-test.cqu.edu.au/adfs/services/trust' because it isn't a valid IdP for this SP.

This patch resolves the issue, in that it correctly sets the IdP in a guest session.

### DIFF
--- a/auth.php
+++ b/auth.php
@@ -458,9 +458,11 @@ class auth_plugin_saml2 extends auth_plugin_base {
         // Fallback in case we can't get the idp from the url param or our session idp is empty.
         // Set the default IdP to be the first in the list. Used when dual login is disabled.
         $fallbackidp = auth_saml2_get_fallback_idp();
-        if (!is_null($fallbackidp)) {
-            $SESSION->saml2idp = md5(auth_saml2_get_fallback_idp()->entityid);
+        if (is_null($fallbackidp)) {
+            // We should already have checked for $fallbackidp === null in the is_configured() function
+            throw new coding_exception('Fallback idp cannot be null in saml_login function');
         }
+        $SESSION->saml2idp = $fallbackidp->entityid;
         $idpfromparam = optional_param('idp', '', PARAM_TEXT);
         if (!empty($idpfromparam)) {
             $SESSION->saml2idp = $idpfromparam;
@@ -484,7 +486,7 @@ class auth_plugin_saml2 extends auth_plugin_base {
             if (!$idpfound) {
                 $this->error_page(get_string('noidpfound', 'auth_saml2', $idpalias));
             }
-        } else if ($idpfromparam) {
+        } else if (!empty($idpfromparam)) {
             $SESSION->saml2idp = $idpfromparam;
         } else if (!is_null($saml2auth->defaultidp)) {
             $SESSION->saml2idp = md5($saml2auth->defaultidp->entityid);
--- a/auth.php
+++ b/auth.php
@@ -455,14 +455,23 @@ class auth_plugin_saml2 extends auth_plugin_base {
         require('setup.php');
         require_once("$CFG->dirroot/login/lib.php");
 
-        // Set the default IdP to be the first in the list. Used when dual login is disabled.
-        $arr = array_reverse($saml2auth->metadataentities);
-        $metadataentities = array_pop($arr);
-        $idpentity = array_pop($metadataentities);
-        $idp = md5($idpentity->entityid);
+        $idpfromparam = optional_param('idp', '', PARAM_TEXT);
+        if (!empty($idpfromparam)) {
+            $SESSION->saml2idp = $idpfromparam;
+        }
 
-        // Specify the default IdP to use.
-        $SESSION->saml2idp = $idp;
+        // Backup in case we can't get the idp from the url param or our session idp is empty.
+        // Set the default IdP to be the first in the list. Used when dual login is disabled.
+        if (!$SESSION->saml2idp) {
+            // Set the default IdP to be the first in the list. Used when dual login is disabled.
+            $arr = array_reverse($saml2auth->metadataentities);
+            $metadataentities = array_pop($arr);
+            $idpentity = array_pop($metadataentities);
+            $idp = $idpentity->entityid;
+
+            // Specify the default IdP to use.
+            $SESSION->saml2idp = $idp;
+        }
 
         // We store the IdP in the session to generate the config/config.php array with the default local SP.
         $idpalias = optional_param('idpalias', '', PARAM_TEXT);
@@ -482,8 +491,6 @@ class auth_plugin_saml2 extends auth_plugin_base {
             if (!$idpfound) {
                 $this->error_page(get_string('noidpfound', 'auth_saml2', $idpalias));
             }
-        } else if (isset($_GET['idp'])) {
-            $SESSION->saml2idp = $_GET['idp'];
         } else if (!is_null($saml2auth->defaultidp)) {
             $SESSION->saml2idp = md5($saml2auth->defaultidp->entityid);
         } else if ($saml2auth->multiidp) {
--- a/auth.php
+++ b/auth.php
@@ -462,15 +462,8 @@ class auth_plugin_saml2 extends auth_plugin_base {
 
         // Backup in case we can't get the idp from the url param or our session idp is empty.
         // Set the default IdP to be the first in the list. Used when dual login is disabled.
-        if (!$SESSION->saml2idp) {
-            // Set the default IdP to be the first in the list. Used when dual login is disabled.
-            $arr = array_reverse($saml2auth->metadataentities);
-            $metadataentities = array_pop($arr);
-            $idpentity = array_pop($metadataentities);
-            $idp = $idpentity->entityid;
-
-            // Specify the default IdP to use.
-            $SESSION->saml2idp = $idp;
+        if (empty($SESSION->saml2idp)) {
+            $SESSION->saml2idp = auth_saml2_get_default_idp();
         }
 
         // We store the IdP in the session to generate the config/config.php array with the default local SP.
@@ -491,9 +484,8 @@ class auth_plugin_saml2 extends auth_plugin_base {
             if (!$idpfound) {
                 $this->error_page(get_string('noidpfound', 'auth_saml2', $idpalias));
             }
-        } else if (!is_null($saml2auth->defaultidp)) {
-            $SESSION->saml2idp = md5($saml2auth->defaultidp->entityid);
-        } else if ($saml2auth->multiidp) {
+        }
+        if ($saml2auth->multiidp) {
             $idpurl = new moodle_url('/auth/saml2/selectidp.php');
             redirect($idpurl);
         }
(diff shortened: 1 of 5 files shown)


dvdcastro:
@kristian-94 is this fix valid for #338?

kristian-94:
@dvdcastro  It does look like it fixes the problem but is still in testing at the moment.

config/authsources.php
```diff
@@ -37,10 +37,9 @@
 $config = [];
 
 // Case for specifying no $SESSION IdP, select the first configured IdP as the default.
-$arr = array_reverse($saml2auth->metadataentities);
-$metadataentities = array_pop($arr);
-$idpentity = array_pop($metadataentities);
-$idp = md5($idpentity->entityid);
+$metadataentities = reset($saml2auth->metadataentities);
+$idpentity = reset($metadataentities);
+$idp = $idpentity->entityid;
```
brendanheywood: While this method works, I'm pretty sure this will fail in scenarios where there are two or more idp's configured and it will always choose the first if the session var is empty and it won't align and you'll get a redirect loop or some sort of failure. I'd like for that scenario to be tested and confirmed one way or another before this lands.

Also I'd like this logic refactored out because it is duplicated here which also needs to be updated:

https://github.com/catalyst/moodle-auth_saml2/blob/master/auth.php#L458-L465

It also seems duplicated by this logic here:

https://github.com/catalyst/moodle-auth_saml2/blob/master/auth.php#L488

It seems like an ideal place for it to go would be here:

https://github.com/catalyst/moodle-auth_saml2/blob/master/locallib.php#L452

Buuuutt ... all of this still feels wrong. If you have 2 idps, and the one you want is not the first, and the session has expired then it will never work. This stuff probably never needs to be in the session in the first place. If it was simply encoded into a query param in the saml returnto url then I think the whole problem is side stepped properly:

https://github.com/catalyst/moodle-auth_saml2/blob/master/auth.php#L507

```
$auth->requireAuth([
    'ReturnTo' => $ME + 'idp=entityid',

```
https://simplesamlphp.org/docs/stable/simplesamlphp-sp-api#section_3

auth.php
```diff
@@ -500,6 +494,11 @@ public function saml_login() {
         $passive = $this->config->duallogin == saml2_settings::OPTION_DUAL_LOGIN_PASSIVE;
         $passive = (bool)optional_param('passive', $passive, PARAM_BOOL);
         $params = ['isPassive' => $passive];
+        // When we return from authenticating, we also return the idp with a url param
+        // This will be found using $_GET['idp'] above
+        $md5idpentityid = auth_saml2_get_md5_idpentityid();
+        $params = ['ReturnTo' => $ME . '?idp=' . $md5idpentityid];
```
brendanheywood: I don't think this will work if $ME has ?foo=bar in it. Please use the moodle_url object to add a param

auth.php
```diff
@@ -500,6 +494,11 @@ public function saml_login() {
         $passive = $this->config->duallogin == saml2_settings::OPTION_DUAL_LOGIN_PASSIVE;
         $passive = (bool)optional_param('passive', $passive, PARAM_BOOL);
         $params = ['isPassive' => $passive];
+        // When we return from authenticating, we also return the idp with a url param
+        // This will be found using $_GET['idp'] above
+        $md5idpentityid = auth_saml2_get_md5_idpentityid();
```
brendanheywood: I think this should simply be $SESSION->saml2idp or you are negating all the logic above

classes/form/selectidp_buttons.php
```diff
@@ -54,6 +54,8 @@ public function definition() {
         $mform->addElement('hidden', 'wants', $wants);
         $mform->addElement('checkbox', 'rememberidp' , '', get_string('rememberidp', 'auth_saml2'));
 
+        $mform->setType('wants', PARAM_RAW);
```
brendanheywood: Can you please also move this up and fix the whitespace so that each form element is grouped in the code

kristian-94:
I've tested this latest patch a bunch, and it works even commenting out these two lines:

https://github.com/catalyst/moodle-auth_saml2/blob/issue338/auth.php#L465

https://github.com/catalyst/moodle-auth_saml2/blob/issue338/config/authsources.php#L58

That means we aren't using the 'default' idp at all: we're getting it from the GET request param. It's still worth leaving this code in as a last resort, and we'll get the first idp from the list if there are multiple. 

The inverse is commenting out the new parts that get the idp from the GET request, ie. 
https://github.com/catalyst/moodle-auth_saml2/blob/issue338/auth.php#L459
https://github.com/catalyst/moodle-auth_saml2/blob/issue338/config/authsources.php#L41

The login still works: that means we're getting the idp from the default idp, the first one in the list. This is the backup scenario where we don't get the idp from the GET request, and I've added comments to indicate that. 

Now just want to get travis passing

auth.php
```diff
@@ -455,14 +455,15 @@ public function saml_login() {
...
-        $idp = md5($idpentity->entityid);
+        if (isset($_GET['idp'])) {
+            $SESSION->saml2idp = $_GET['idp'];
+        }
 
-        // Specify the default IdP to use.
-        $SESSION->saml2idp = $idp;
+        // Backup in case we can't get the idp from the url param or our session idp is empty.
+        // Set the default IdP to be the first in the list. Used when dual login is disabled.
+        if (!isset($SESSION->saml2idp) || (isset($SESSION->saml2idp) && !$SESSION->saml2idp)) {
```
brendanheywood: This seems overly complicated, could you use !empty() ?

kristian-94: Yes, I've updated that

locallib.php
```diff
@@ -450,7 +450,7 @@ function auth_saml2_get_idps($active = false, $asarray = false) {
  * @return object The default IdP object, or NULL if there is no default IdP set.
```
brendanheywood: this phpdoc needs a slight tweak now

Can you please wrap this in unit tests. I'm not sure this is correct, it appears in some cases to return an object, and in other cases to return an entityid

kristian-94: Hi @brendanheywood , what would you suggest for the phpdoc? It's still possible for that function to return null if we have no idp's set at all.

kristian-94: I have added a unit test, and fixed that function and the place in config/authsources.php where we need to get the entity id from the returned object. https://github.com/catalyst/moodle-auth_saml2/pull/344/files#diff-36bb225ac3d28fa53bb66e657ddefdcaR58

auth.php
```diff
@@ -455,14 +455,15 @@ public function saml_login() {
         require('setup.php');
         require_once("$CFG->dirroot/login/lib.php");
 
-        // Set the default IdP to be the first in the list. Used when dual login is disabled.
-        $arr = array_reverse($saml2auth->metadataentities);
-        $metadataentities = array_pop($arr);
-        $idpentity = array_pop($metadataentities);
-        $idp = md5($idpentity->entityid);
+        if (isset($_GET['idp'])) {
```
brendanheywood: I know this is moved code, but any reason for not using optional_param ?

classes/form/selectidp_buttons.php
brendanheywood: Please move this up to where it should be

config/authsources.php
```diff
@@ -36,11 +36,10 @@
...
 $config = [];
 
-// Case for specifying no $SESSION IdP, select the first configured IdP as the default.
-$arr = array_reverse($saml2auth->metadataentities);
-$metadataentities = array_pop($arr);
-$idpentity = array_pop($metadataentities);
-$idp = md5($idpentity->entityid);
+// Grab the idp from the request if it's available
+if (isset($_GET['idp'])) {
+    $SESSION->saml2idp = $_GET['idp'];
```
brendanheywood: Is this supposed to be setting $SESSION->saml2idp  or $idp?

auth.php
kristian-94: No reason, let's use that instead, have updated that now

classes/form/selectidp_buttons.php
kristian-94: Have added this now in its own commit

config/authsources.php
kristian-94: You're right, it's meant to be $idp here, so it is set later in the $config variable

brendanheywood:
There is a regressions here when multiple idp's are setup I keep geting pushed to:

http://moodle.local/auth/saml2/selectidp.php

and I can't get past that page

locallib.php
```diff
@@ -439,18 +439,37 @@ function auth_saml2_get_idps($active = false, $asarray = false) {
...
+    }
+    return $fallbackidp;
+}
 /**
  * This helper function returns the default IdP if it is configured.
  * @return object The default IdP object, or NULL if there is no default IdP set.
  */
 function auth_saml2_get_default_idp() {
-    global $DB;
+    global $DB, $saml2auth;
```
brendanheywood: typo? can you remove this?

tests/phpunit/locallib_test.php
```diff
@@ -444,5 +444,46 @@ public function test_is_email_taken() {
...
     }
+    public function test_get_fallback_idp_returns_idp_object() {
+        global $DB, $saml2auth;
+        $this->resetAfterTest();
+        // First test when we have no idp's configured.
+        $fallbackidp = auth_saml2_get_fallback_idp();
+        $this->assertTrue(is_null($fallbackidp));
+
+        // Add two fake IdPs.
+        $metadataurl = 'https://idp.example.org/idp/shibboleth';
```
brendanheywood: Instead of inserting these, can you just set the config via the settings object so we are properly testing how it will be done for real?

config/authsources.php
```diff
@@ -36,11 +36,10 @@
 
 $config = [];
 
-// Case for specifying no $SESSION IdP, select the first configured IdP as the default.
-$arr = array_reverse($saml2auth->metadataentities);
-$metadataentities = array_pop($arr);
-$idpentity = array_pop($metadataentities);
-$idp = md5($idpentity->entityid);
+// Grab the idp from the request if it's available
+if (isset($_GET['idp'])) {
```
brendanheywood: $_GET vs optional_param?

auth.php
```diff
@@ -455,14 +455,16 @@ public function saml_login() {
...
-        $arr = array_reverse($saml2auth->metadataentities);
-        $metadataentities = array_pop($arr);
-        $idpentity = array_pop($metadataentities);
-        $idp = md5($idpentity->entityid);
-
-        // Specify the default IdP to use.
-        $SESSION->saml2idp = $idp;
+        $fallbackidp = auth_saml2_get_fallback_idp();
+        if (!is_null($fallbackidp)) {
+            $SESSION->saml2idp = md5(auth_saml2_get_fallback_idp()->entityid);
```
brendanheywood: use $fallbackidp as above instead?

kristian-94:
This looks fixed in master now, and the original issue #338 is closed, so we can close this pr.

PR #349, opened 2019-08-26
Fix #348 - allow private cert key to be configurable.

### LINKED ISSUE #348: allow a different var to be used for privatekey other than siteidentifier.
saml2 uses the Moodle siteidentifier as a key for encrypting the certs. Moodle is using the siteidentifier a bit more these days - in particular for site registration which expects the siteidentifier and url to be a unique combination.

This is a bigger issue when a site wants to try using notifications on a testing site with the Moodle app which requires registration. If they sync their prod database into a testing site, then try to register the site it will find that another site with that id is registered with a different url so it will reset the local siteid, which will then break the saml2 auth on the testing site because it cannot decrypt the cert.

I'm thinking we should just allow a config var to set a different value to be used - maybe something like:
$CFG->forced_plugin_settings = array('auth_saml2'  => array('certkey' => 'somehash'),

I don't think we need a UI to manage this but let me know what you think.

danmarsden:
Allows the same cert to be used in testing/prod sites with different SITE ID's


danmarsden:
pinging @brendanheywood - this look ok as a solution here?

settings.php
```diff
@@ -142,6 +142,13 @@
             $CFG->wwwroot . '/auth/saml2/regenerate.php'
             ));
 
+    $settings->add(new admin_setting_configtext(
```
brendanheywood: would this be better as admin_setting_configpasswordunmask ?

settings.php
```diff
@@ -142,6 +142,13 @@
...
+        'auth_saml2/privatekeypass',
+        get_string('privatekeypass', 'auth_saml2'),
+        get_string('privatekeypass_help', 'auth_saml2'),
+        get_site_identifier(),
```
brendanheywood: I'm wondering wether it would be better if this defaulted to an empty string, and if it was empty then it dynamically defaults back to get_site_identifier()?

I'm thinking around when data gets refreshed to other environments at the moment this should correctly break, but with this patch it will now work. Maybe this is a feature, maybe this is a bug. From a security perspective I'd prefer a refresh script to just force set the value of the privatekeypass and not inherit it and leak a private key

db/upgrade.php
```diff
@@ -280,5 +280,11 @@ function xmldb_auth_saml2_upgrade($oldversion) {
         upgrade_plugin_savepoint(true, 2019022100, 'auth', 'saml2');
     }
 
+    if ($oldversion < 2019062600) {
+        // Move private key into new setting.
+        set_config('privatekeypass', 'auth_saml2');
```
brendanheywood: Not sure this looks correct, shouldn't it set the value to get_site_identifier()?

But first lets resolve the other question below, this could be redundant

settings.php
danmarsden: If the privatekeypass is supposed to be "secret" I'm not convinced we should be using the site identifier here at all? - Most uses of the site_identifer (backup xml etc) use an MD5 of the siteidentifier, but that's not guaranteed I think. The existing use of site identifier is why I implemented this as a setting in the end - rather than just a hard-coded value set at the db level.

If it should be secret we could create a hash stored in the config table on install (with no UI) that is only used by the saml private key instead of relying on the site identifier. (existing installs would just copy the existing site identifier into this new config var.)

Also - when data gets "refreshed" from a prod to testing site the siteidentifier is usually pulled across from the prod database at the same time, (unless using some form of plugin for transferring data instead of a db level copy.) the Moodle site doesn't regenerate the siteidentifier unless you trigger a hub registration and it finds a url and siteidentifier don't match.

that sound any better?

lang/en/auth_saml2.php
```diff
@@ -105,6 +105,8 @@
 $string['organizationname'] = 'Organisation';
 $string['passivemode'] = 'Passive mode';
 $string['pluginname'] = 'SAML2';
+$string['privatekeypass'] = 'Private certificate key';
```
brendanheywood: I think a slight tweak, this is not the cert key, it's the cert key password.

version.php
```diff
@@ -24,7 +24,7 @@
 
 defined('MOODLE_INTERNAL') || die();
 
-$plugin->version   = 2019061200;    // The current plugin version (Date: YYYYMMDDXX).
+$plugin->version   = 2019062600;    // The current plugin version (Date: YYYYMMDDXX).
 $plugin->release   = 2019061200;    // Match release exactly to version.
```
brendanheywood: Also can you please bump $plugin-release in lockstop with plugin->version

settings.php
brendanheywood: I just double checked and I was wrong, the site id isn't reset from a 'raw' refresh, this is done in the local_datacleaner plugin. For some reason I thought if the domain changed the siteid got reset.

I don't think this config item is any more sensitive than some other critical admin settings, like oauth tokens etc so I don't think it's worth implementing a special admin setting for this, but I do think swapping it to use admin_setting_configpasswordunmask is enough and that's the main thing I'd like to see to land. That opens the door for some tighter security across the board for those that need it via MDL-66445 which we'll tackle shortly

lang/en/auth_saml2.php
danmarsden: yeah - good poiint - done.

version.php
danmarsden: whoops - missed you were doing that.

settings.php
danmarsden: cool - will be nice to see MDL-66445 land at some point. I've fixed those and rebased for you to take another look.

Issue #194, opened 2018-02-20
"Allowed any auth type" exhibits strange timeout


egiles:
**What happened?**
When using the 'Allowed any auth type' option, for users who's auth type is 'ldap'..... they are being logged out after a couple of minutes of inactivity.

**What I expected:**
Users should remain logged in until session timeouts kick in (longer than two minutes!)

**How to reproduce:**

- Create a user whose auth type is 'ldap'
- Configure SAML plugin:  Dual login = Yes
- Configure SAML plugin:  Allowed any auth type = Yes
- Login as the ldap user (but using SAML)
- browse to /admin/user.php
- enter some text in the search field (don't hit enter!)
- wait two minutes
- click add filter
- you're now logged out!

** Thoughts. **
This is quite possibly a core Moodle/Totara bug.  I am using Totara 2.9

brendanheywood:
I think the root cause here is if you are say a 'OAuth2' user and you login via saml, you aren't fully perfectly logged in and a timestamp isn't being touched so after the  \core\task\session_cleanup_task runs your session gets gc()'ed. 

This may be something specific to OAuth2 users

brendanheywood:
A quick workaround is to change the session cleanup task from every 5 mins to once each night. It still needs to run, and the front end sessions will still expire based on the cookie value and the session timeout. YMMV.

egiles:
Thanks @brendanheywood.

Actually, our workaround was to update all users so they had an authentication type of saml2.  In this case that seems to be the right thing to do (since our users were all logging in with saml2).

I am happy to close this ticket if that's all we can do? I only created the ticket in the hope that there is a way to make the code handle this situation in a more robust manner.

brendanheywood:
I'd still like to leave it open and get to the bottom of it. I'm quite sure what that means yet. The problem with changing type is for people who are provisioned using some other means, eg the ldap sync plugin will be ldap users

brendanheywood:
The 'ignore_timeout_hook` may come into play here:

https://github.com/moodle/moodle/blob/master/lib/classes/session/manager.php#L784

Issue #42, opened 2016-08-09
Rule out session age bug with cron cleanup


brendanheywood:
From plugin page:

> We are using the setting Allow any auth types and importing our users into imap and ldap (for several reason) We also had these auth types disabled just to be sure all users must come in through saml
>
> Discovered that the Moodle task Cleanup old sessions would then remove all sessions and we would get new sessions every cron which will cause you to get invalidsesskey if it runs while you are in the middle of forum post or another process
>
> Its debatable if this is a bug as it should only affect saml2 under specific settings

brendanheywood:
I have a vague feeling this may be due to the logged in time, and last login, timemodified timestamps not being set correctly. It should be easy to reproduce by setting the $CFG->sessiontimeout to something super small like 1 minute and then running the cleanup task:

```
php admin/tool/task/cli/schedule_task.php --execute='\core\task\session_cleanup_task'
```

This may also be an issue with auth_basic and others, so if you can reproduce this and fix it lets side port it to the other auth plugins as needed.

nhoobin:
Keeping track of a few details that we discussed regarding the plugin page comment.

The session cleanup task will remove the sessions of plugins that have been disabled, which may be why the user was reporting that sessions were being regenerated and the sesskey was invalid. https://github.com/moodle/moodle/blob/master/lib/classes/session/manager.php#L744

During debugging we also found that it may be possible to implement an auth plugin hook ignore_timeout_hook to prevent the sessions from being removed based on how we code it. A basic implementation is on this commit https://github.com/nhoobin/moodle-auth_saml2/commit/647174cbc12a51d179c29302f88fd1d88c96efd1

Yet with the discovery of that hook we identified that core Moodle does not properly respect the return value for the function ignore_timeout_hook and we created a tracker ticket to resolve this. https://tracker.moodle.org/browse/MDL-56417

PR #358, opened 2019-09-25
Add warning to admin if the site has disabled authentication plugin

danmarsden:
Basically when the allow any users setting is enabled, it allows any user in the system to login - even if they have a "disabled" authentication plugin in their profile.

Moodle's garbage collection of sessions deletes the sessions of users with a disabled authentication plugin - (typically every few minutes) so these users have to login to Moodle which might happen quietly behind the scenes because they are already authenticated to the IDP, but causes uncessary load on the server.

This just adds a big warning on the settings page if it detects the setting is enabled and if it can find any users in the site with a disabled auth - hopefully using record_exists is pretty quick/cheap.

@brendanheywood - this look ok to you?

The internal moodle session clean up tasks deletes sessions for users with disabled auth types.

### DIFF
--- a/lang/en/auth_saml2.php
+++ b/lang/en/auth_saml2.php
@@ -48,6 +48,7 @@ $string['commonname'] = 'Common Name';
 $string['countryname'] = 'Country';
 $string['debug'] = 'Debugging';
 $string['debug_help'] = '<p>This adds extra debugging to the normal moodle log | <a href=\'{$a}\'>View SSP config</a></p>';
+$string['auth_saml2disabledauthusers'] = '<p class="alert alert-warning">The ability for any auth type to use the SAML Login has been enabled, but there are users in your site linked to a disabled authentication plugin. These users may be able to login, but the Moodle session clean up task will delete their session, requiring them to login again within a short time period. Please delete these users or update the authentication used in their profile.</p>';
 $string['duallogin'] = 'Dual login';
 $string['duallogin_help'] = '
 <p>If on, then users will see both manual and a SAML login button. If off they will always be taken directly to the IdP login page.</p>
--- a/settings.php
+++ b/settings.php
@@ -43,6 +43,29 @@ if ($ADMIN->fulltree) {
     $settings->add(new admin_setting_heading('auth_saml2/pluginname', '',
         new lang_string('auth_saml2description', 'auth_saml2')));
 
+    if (!empty(get_config("auth_saml2", "anyauth")) && strpos($CFG->auth, 'saml2')) {
+        // Check for disabled auth plugins that have users.
+        $disabledplugins = array();
+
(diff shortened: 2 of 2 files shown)


brendanheywood:
Agree with the issue but something feels wrong with the solution. Typically the main reason you want this feature is:

a) setup and sync via ldap, but auth via saml. This is a workaround a deep structural issue with moodle not treating identity as a separate concept to authentication which will continue to haunt us forever. In this case the ldap auth should definitely be enabled. In this case if SSO breaks we can very easily flip a bit and ldap just works again.

b) legacy / migration management, ie we were using auth_xyz, now we are using SSO but don't want to migrate the users, typically because of some other integration that might break. In this case we still want the users auth plugin to be enabled.

I'm struggling to think of a use case where leveraging a user who is in an auth type which is disabled, but given that it is happening should be instead implement ignore_timeout_hook instead?:

https://github.com/moodle/moodle/blob/master/lib/classes/session/manager.php#L359

This is likely the root cause or at least related to these:

https://github.com/catalyst/moodle-auth_saml2/issues/42 Rule out session age bug with cron cleanup

https://github.com/catalyst/moodle-auth_saml2/issues/194 "Allowed any auth type" exhibits strange logout timeout

So logic something like:

If we logged in via saml, set a bit in their session. If and only if that bit is set, then ignore the session timeout in the cron task, but still honor a 'normal session timeout as if they were a saml auth type user.

I think that admin warning is great, but it's still a crappy UX for the students who get booted out and the admin may never see that message. I think these are complimentary.

danmarsden:
Personally I'm not sure the allow any user setting should really work for users with a disabled auth plugin but the fact that it does now makes it hard to fix properly. I think the core behavior of clearing user sessions for users with a disabled plugin seems to make sense.

danmarsden:
Also - ignore timeout check only applies to enabled auth plugins in core from what I could see...

brendanheywood:
It's either valid, in which case to don't log them out after a minute, or it is not valid, in which case we should not allow them login in the first place. I'm fine with either approach but not the middle ground which is confusing and non deterministic, you should get a clean error as a user.

brendanheywood:
>  ignore timeout check only applies to enabled auth plugins in core from what I could see

that seals the deal then, I think we must not let them login in the first place

brendanheywood:
@danmarsden just having reviewed the other patch around not allowing duplicate emails, it does feel very similar to this one. ie I'm firmly in the camp of not letting them log in at all

danmarsden:
Yep - agree, we should just prevent login. I'll try to come back to this at some point.

brendanheywood:
@dan I've implemented this so I'll close this pr

https://github.com/catalyst/moodle-auth_saml2/commit/0df8b4206cdf57f47fbdf4917c284b5a810f3730

PR #360, opened 2019-10-02
Improved actions for blocked users

golenkovm:


auth.php
```diff
@@ -60,6 +60,11 @@ class auth_plugin_saml2 extends auth_plugin_base {
         'logtofile'          => 0,
         'logdir'             => '/tmp/',
         'nameidasattrib'     => 0,
+        'flagresponsetype'   => saml2_settings::OPTION_FLAGGED_LOGIN_NONE,
+        'flagattribute'      => 'encumbered',
```
dmitriim: I think  don't need this anymore as well as flagvalue. Also related strings need to be deleted as well.

auth.php
```diff
@@ -525,7 +531,8 @@ public function saml_login() {
 
         // Testing user's groups and allow access decided on preferences
         if (!$this->is_access_allowed_for_member($attributes)) {
-            $this->error_page(get_string('wrongauth', 'auth_saml2', $uid));
+            //$this->error_page(get_string('wrongauth', 'auth_saml2', $uid));
```
dmitriim: Maybe remove this commented out code?

auth.php
```diff
@@ -602,25 +609,88 @@ public function saml_login() {
...
             return false;
         } else {
             $this->log(__FUNCTION__ . " user '$uid' is in allowed group and isn't in restricted. Access allowed.");
+            return true;
         }
+
+        // If a user has an encumberance attribute and one of the groups in it match a deny group,
+        // then don't let them in.
+        // We realise that they may not get here if they are in an allow group.
+        if (!empty(array_intersect($deny, $groups))) {
```
dmitriim: it seems like this code will never be reached.

classes/admin/saml2_settings.php
```diff
@@ -28,6 +28,7 @@
 /**
  * @package     auth_saml2
  * @author      Daniel Thee Roperto <daniel.roperto@catalyst-au.net>
+ * @link        http://tools.ietf.org/html/rfc3986 (the URI specification)
```
dmitriim: Not sure if we nee this link

tests/phpunit/settings_helper_test.php
```diff
@@ -0,0 +1,81 @@
...
+        return $validurls;
+    }
+
+    /**
+     * Provider for invalid urls from fixture
+     *
+     * @return mixed
+     */
+    public function provide_invalid_urls() {
+        require_once(__DIR__ . '/../fixtures/invalidurls.php');
```
brendanheywood: This should be a normal data provider

classes/settings_helper.php
```diff
@@ -0,0 +1,69 @@
...
+    /**
+     * @var string representation of PCRE Regular Expression for checking http/https scheme URLs
+     * in accordance with RFC3986.
+     */
+    private static $httpsregex;
+
+    /**
+     * Setter for $http_https_regex.
+     */
+    protected static function set_https_regex() {
```
brendanheywood: The setter and setting here are overkill, just make this a public attribute, or a getter with the regex inline.

tests/phpunit/settings_helper_test.php
```diff
@@ -0,0 +1,81 @@
...
+defined('MOODLE_INTERNAL') || die();
+
+require_once(__DIR__ . '/../../_autoload.php');
+
+/**
+ * Class settings_helper_testcase
+ *
+ * @package auth_saml2\tests
+ */
+class settings_helper_testcase extends advanced_testcase {
```
brendanheywood: Why does any of this file exist at all? Moodle has internal libs for parsing and validating urls, shouldn't we just use the,?

dmitriim: @tomdickman  should know the exact answer as he wrote this code. It's related to saml library being very picky about URL and moodle API is not enough strict

brendanheywood:
I don't want this to land. I've had this conversation multiple times with different and I don't understand why we keep wasting time going in circles. If this was done as a separate plugin the code would be simpler, and re-usable with other auth types.

dmitriim:
@brendanheywood this patch doesn't introduce any conceptually new functionality. It extends existing group restriction to  make it a bit more configurable. In other words replaces hardcoded error_page to a bit more flexible actions like redirect or error_page with a configurable message.

tests/phpunit/settings_helper_test.php
dmitriim: fixed.

classes/settings_helper.php
dmitriim: simplified this code and move this to a constant of saml2_settings class

auth.php
```diff
@@ -609,27 +613,78 @@ public function saml_login() {
...
 
+    /**
+     * Redirect SAML2 login if a flagredirecturl has been configured.
+     *
+     * @throws \moodle_exception
+     */
+    protected function redirect_blocked_access() {
+
+        if (!empty($this->config->flagredirecturl)) {
+            redirect(new moodle_url($this->config->flagredirecturl));
```
mattporritt: Does `moodle_url` handle non Moodle URLS? I thought we wanted to be able to redirect users to any valid URL?

classes/admin/saml2_settings.php
```diff
@@ -41,4 +41,18 @@ abstract class saml2_settings {
...
+
+    const OPTION_FLAGGED_LOGIN_MESSAGE = 1;
+
+    const OPTION_FLAGGED_LOGIN_REDIRECT = 2;
+
+    /**
+     * @var string representation of PCRE Regular Expression for checking http/https scheme URLs
+     * in accordance with RFC3986.
+     */
+    const URL_HTTPS_REGEX = '/(^(https?\\:\\/\\/(www\\.)?'
```
mattporritt: This is gross. I read the comment about SAML lib being more picky than Moodle lib, but I still don't understand why we need it. It appears it's Moodle core that processes this, or did I miss something?

auth.php
dmitriim: Yeah, it could be any URL not just moodle.

classes/admin/saml2_settings.php
dmitriim: Actually it could be a leftover from the initial implementation. I'll check the code. @tomdickman can you please remind us why it was done in this way?

auth.php
tomdickman: `moodle_url` doesn't have to be internal, it can be any url if it's a fully qualified domain name, or an internal url if it's relative.

classes/admin/saml2_settings.php
tomdickman: Hey @mattporritt  and @dmitriim, we had a discussion about this originally as I didn't really want to have to do it and instead I wanted to use the Moodle inbuilt form param types for adding the setting, However, in testing none of the param constants worked for URLs as expected. Specifically `PARAM_URL` still let through poorly formed URLs (in fact it didn't seem to parse anything at all if I remember correctly, it's been a while).
As a result the decision was made to use regex and just add some extensive test coverage, as a workaround.

dmitriim: @mattporritt it's up to you to keep this code or I can just use PARAM_URL and leave it to admins to make sure that not screw it up.

tests/phpunit/settings_helper_test.php
dmitriim: Replaced with PARAM_URL and removed unused code.

classes/admin/saml2_settings.php
dmitriim: Replaced with PARAM_URL and removed unused code.

PR #365, opened 2019-10-15
Respect allowaccountssameemail setting

dmitriim:


auth.php
```diff
@@ -697,6 +715,57 @@ public function update_user_profile_fields(&$user, $attributes, $newuser = false
...
 
+    /**
+     * Get email address from attributes.
+     *
+     * @param array $attributes A list of attributes.
+     *
+     * @return bool
+     */
+    public function get_email_from_attributes(array $attributes) {
+        $emailattr = get_config('auth_saml2', 'field_map_email');
```
brendanheywood: why not use $this->config ?

dmitriim: I have copied it from https://github.com/catalyst/moodle-auth_saml2/blob/master/auth.php#L659 But you are right we should user $this->config here. Will fix it.

dmitriim: Refactored the code to use $this->config

brendanheywood:
This is good to go,just waiting for travis

Issue #369, opened 2019-11-04
upcoming security release for simplesamlphp


danmarsden:
https://groups.google.com/forum/#!topic/simplesamlphp-announce/2odMqz63z7k

we'll need to update our versions pretty quick too.

jtsafran:
@danmarsden I was just coming here to report this.  Glad you all are on top of it!

If you need any help testing, please let me know.  I have a few dev/test sites that are configured against ADFS, Okta, and WSO2 (Ellucian Ethos Identity rebranded) that we can test against.  

Thank you for the awesome plugin and being on top of this coming SSPHP bug!

danmarsden:
Feel free to grab that existing branch and give any feedback on testing/ any errors you see - we are seeing some php lint issues with the symfony framework but I've got it running on one site so far.

Also need to work out why those behat failures are occuring....

danmarsden:
latest simplesaml-1.17.6-v4 branch is now passing tests - we need to test this updated branch in a few more sites before we merge into master - let us know if you get a chance to do this!

All going well, we'll then apply the security patch on top when it is released.

Richardvi:
Thanks for picking this up, was a bit worried...

jtsafran:
@danmarsden Downloading and installing it, now.  Will have a follow up in a few hours!

danmarsden:
simplesaml-1.17.7 branch here in github includes the latest branch (based on master) with the upgrade to 17.7 that includes the security fix.

I haven't had a chance to test this and I'm out of the office this morning, but someone else in the team might test and merge it in.

jtsafran:
Dan,

1.17.6 tests went well.  No issues that I am seeing as far as functionality.  The only thing that looked off is the SSPHP version was showing as "master" in the plugin.  Isn't the master branch of SSPHP the 18 RC?

I'll pull down and test functionality of the 1.17.7 version shortly.

-Jesse

danmarsden:
1.17.7 now merged into Master (Moodle 3.5 and higher, Totara 12 and higher) - this has been rolled out to a few sites and no adverse issues seen so far.

new patch in testing for the older Totara branches in the 27_34-xmlsec branch - not tested very widely yet.

danmarsden:
27_34 branch now merged in as well - tested ok in a few sites, and minor change to this branch, just updating the xmlseclib - will be interesting to see the full details when the embargo is over!

PR #370, opened 2019-11-04
Simplesaml 1.17.6 (don't merge yet)

danmarsden:
update simplesaml to 1.17.6 to help prepare for #369


danmarsden:
few notes on this..

Deployed this to a testing site and the updated code seems to still work for an exisiting IDP.

Issues
1) comes bundled with a bunch of polyfills for 7.0/7.2 - we lint all files prior to deployment (and in travis) and these files fail the lint - we need to tidy up the files to remove these.

2) Behat tests failing - not sure why, but doesn't look like the php lint issues are causing that so it's likely to need further investigation.

peterspicer-catalyst:
@danmarsden would it be easier to remove the polyfill from the composer file then rerun composer than trying to manually remove the polyfills from the resultant autoloader? Might make our maintenance easier in the long run and be cleaner about weird random dependencies.

danmarsden:
@peterspicer-catalyst - yeah I was playing around a bit with that yesterday, looks like we're going to have some other challenges with the symfony lib version included in simplesamlphp - not just with the php-70 polyfill.

Will also be interesting to see if we can backport the security fix when it is released to the older saml2 branches for Totara support.

I guess I'll play around with this a bit more today...

danmarsden:
further update - I just created a new branch with updated composer dependencies for php 7.2 to see if that helps - still has a few issues unfortunately!

- https://travis-ci.org/catalyst/moodle-auth_saml2/jobs/607868155#L2305

I'm tempted to look at using the simplesamlphp 1.18 RC to see if it helps us...

danmarsden:
looks like behat failures relate to the "passive" login feature which isn't working with the latest update.

danmarsden:
I'm also leaning towards not updating all the dependencies and just removing the stuff that fails lint.

danmarsden:
closing this one as I have a better wip branch.

PR #377, opened 2019-11-07
Add a new setting Allowed Groups Priority to close #375

### LINKED ISSUE #375: New setting to control order of applying restricted_groups and allowed_groups
At the moment if a user is in restricted_groups and allowed_groups at the same time he will be blocked from login. 
In our case we would like to allow the user in if he is in both groups.

dmitriim:
#375

### DIFF
--- a/settings.php
+++ b/settings.php
@@ -258,6 +258,13 @@ if ($ADMIN->fulltree) {
             'student',
             PARAM_TEXT));
 
+    // Priority of the allowed groups.
+    $settings->add(new admin_setting_configselect(
+        'auth_saml2/allowedgroupspriority',
+        get_string('allowedgroupspriority', 'auth_saml2'),
+        get_string('allowedgroupspriority_help', 'auth_saml2'),
+        0, $yesno));
+
     // Alternative Logout URL.
     $settings->add(new admin_setting_configtext(
             'auth_saml2/alterlogout',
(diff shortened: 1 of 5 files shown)


settings.php
```diff
@@ -258,6 +258,13 @@
             'student',
             PARAM_TEXT));
 
+    // Priority of the allowed groups.
```
brendanheywood: Putting aside the whole 'should this exist at all' having 3 different admin settings bloated and hard to see what is going on. We now have a list of restricted groups:

```
group1
group2
```
and a list of allowed groups:
```
group2
group3
```

and now we have this 3rd option which says which order to process those ACL rules in. If it is true then allowedgroups are always allowed ie (group2 = yes). If it is false then if they are in restricted then they are always denied (group2 is denied). If they are in neither group then they are always allowed (group4 is allowed). Inevitable someone is going to ask for some more complex business logic again.

We are slowly recreating an ACL language. We now have 4 settings for an edge case feature that very few people will ever use, it's a lot of clutter. Why not just ditch all 3 settings and just have one setting like this:

```
allow group1
allow group2
deny group3
```

These ACL's would be resolved in the order they are written so by swapping the order you can satisfy the business problem of having higher or lower precedence ACL's as well as making it clearer at a glance what will actually happen. It also allows the more complex business rules where group1 should always be allowed, but group3 deny directive overrides group2's allow directive. The config migration path between the old 2 settings and the new 1 is trivial, just prefix them and tack them together.

Even better we can merge all 4 settings and at the same time it even more powerful and ditch the original groupattr setting by doing this:

```
allow group=group1
allow group=group2
deny group=group3
```

Because then you could leverage different attributes together:

```
allow admins=member
deny thieves=member
allow examrole=proctor
deny library=overdue
```

Again the only small change in the migration is prefixing groupattr= to each group.

dmitriim: good idea

dmitriim:
I'll close this one as I can't change the branch. A new PR is up https://github.com/catalyst/moodle-auth_saml2/pull/450

Issue #388, opened 2020-01-15
Add an callback for auth proc filters at moodle level from other plugins


brendanheywood:
**What happened?**

SSP has an authproc concept:

https://simplesamlphp.org/docs/stable/simplesamlphp-authproc

But there is no easy way to use without either forking the plugin or modifying the code etc. So proposing to introduce a new moodle level callback, which *any* plugin can implement, which returns a link of auth proc filters. Saml will call them all, and load them into the sp config and then let SSP handle them as it normally would.

Then the complete suite of filters is available in a way which is very simple to implement in a clean way:

https://simplesamlphp.org/docs/stable/simplesamlphp-authproc#section_2

* [ ] implement new plugin callback
* [ ] pull that into the sp conf
* [ ] readme with full example (eg core:PHP)

brendanheywood:
Standard pattern for callbacks:

```
lib/navigationlib.php-2551-            // Let plugins hook into user navigation.
lib/navigationlib.php:2552:            $pluginsfunction = get_plugins_with_function('extend_navigation_user', 'lib.php');
lib/navigationlib.php-2553-            foreach ($pluginsfunction as $plugintype => $plugins) {
lib/navigationlib.php-2554-                if ($plugintype != 'report') {
lib/navigationlib.php-2555-                    foreach ($plugins as $pluginfunction) {
lib/navigationlib.php-2556-                        $pluginfunction($usernode, $user, $usercontext, $course, $coursecontext);
lib/navigationlib.php-2557-                    }
lib/navigationlib.php-2558-                }
lib/navigationlib.php-2559-            }
```

call the callback something that will never clash with another core callback like:


get_plugins_with_function('extend_auth_saml2_proc', 'lib.php');

and then in say local_xyz / lib.php you would implement:

```
function local_xyz_extend_auth_saml2_proc() {
   return [
        51 => array(
            'class' => 'core:AttributeMap',
            'oid2name',
        ),
        52 => array(
            'class' => 'core:AttributeMap',
            'oid2name',
        ),
    ]
}
```

PR #390, opened 2020-02-12
Add auth proc filters hook

gmrsagar:
Addressing the issue #388

### DIFF
--- a/README.md
+++ b/README.md
@@ -276,6 +276,33 @@ Suggested attribute mappings:
 |`Email`|`user.email`|
 
 
+**Auth Proc Filter Hooks**
+
+Other plugins may hook into SAML2 and create custom Auth Proc Filters.
+Auth Proc Filters allows you to do stuff at the IdP after authentication is complete
+and just before you are sent back to the SP
+
+Steps to implement the hook:
+* Create a plugin that will implement the hook (e.g local_hookimplement)
+* Define the hook function 'local_hookimplement_extend_auth_saml2_proc' in plugin's lib.php
+* The function should return array of SimpleSaml Auth Proc Filters.
+
+Example:
+```php
+function local_hookimplement_extend_auth_saml2_proc {
+   // Do your data validation
+
+   return [
+      51 => array(
+         'class' => 'core:AttributeMap',
+         'oid2name'
+      )
+   ]
+}
+```
+The function should return an array of Filters that is to be applied. The index of the array represents the priority for the filters.
+
+
 
 Other SAML plugins
 ------------------
--- a/auth.php
+++ b/auth.php
@@ -104,6 +104,37 @@ class auth_plugin_saml2 extends auth_plugin_base {
         $this->defaultidp = auth_saml2_get_default_idp();
     }
 
+    public static function saml2_authproc_filters_hook() {
+        $authprocfilters = [];
+        $authprocfilters[50] = array(
+            'class' => 'core:AttributeMap',
+            'oid2name',
+        );
+        $callbacks = get_plugins_with_function('extend_auth_saml2_proc', 'lib.php');
+        foreach ($callbacks as $plugins) {
+            foreach ($plugins as $pluginfunction) {
+                $filters = $pluginfunction();
+                foreach ($filters as $key => $value) {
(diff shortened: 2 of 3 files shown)


config/config.php
```diff
@@ -88,12 +88,7 @@
 
     'proxy' => null, // TODO inherit from moodle conf see http://moodle.local/admin/settings.php?section=http for more.
 
-    'authproc.sp' => array(
-        50 => array(
-            'class' => 'core:AttributeMap',
-            'oid2name',
-        ),
-    ),
+    'authproc.sp' => auth_plugin_saml2::saml2_authproc_filters_hook(),
```
brendanheywood: look good, but please don't remove the core:AttributeMap one that was already there. Or move this into saml2_authproc_filters_hook

auth.php
```diff
@@ -104,6 +104,19 @@ public function __construct() {
         $this->defaultidp = auth_saml2_get_default_idp();
     }
 
+    public static function saml2_authproc_filters_hook()
+    {
+        if (function_exists('get_plugins_with_function')) {
```
brendanheywood: is this if needed?

auth.php
```diff
@@ -104,6 +104,19 @@ public function __construct() {
...
+            $callbacks = get_plugins_with_function('extend_auth_saml2_proc', 'lib.php');
+            foreach ($callbacks as $type => $plugins) {
+                foreach ($plugins as $plugin => $pluginfunction) {
+                    $authprocfilters = $pluginfunction();
```
brendanheywood: I think you want to keep appending more items onto the array, and also start the array with an explicit empty array

config/config.php
gmrsagar: moved this to saml2_authproc_filters_hook as a default hook

auth.php
gmrsagar: you're right this is not needed. removed it.

auth.php
gmrsagar: updated with a default filter.

auth.php
```diff
@@ -104,6 +104,37 @@ public function __construct() {
...
+            'class' => 'core:AttributeMap',
+            'oid2name',
+        );
+        $callbacks = get_plugins_with_function('extend_auth_saml2_proc', 'lib.php');
+        foreach ($callbacks as $plugins) {
+            foreach ($plugins as $pluginfunction) {
+                $filters = $pluginfunction();
+                foreach ($filters as $key => $value) {
+                    $key = self::check_filters_priority($key, $authprocfilters);
+                    $authprocfilters[$key] = $value;
```
brendanheywood: can you please emit a debugger warning when two proc filters have the same key and clash

README.md
```diff
@@ -276,6 +276,33 @@ Suggested attribute mappings:
 |`Email`|`user.email`|
 
 
+**Auth Proc Filter Hooks**
+
+Other plugins may hook into SAML2 and create custom Auth Proc Filters.
+Auth Proc Filters allows you to do stuff at the IdP after authentication is complete
```
brendanheywood: This isn't quite correct

"Auth Proc Filters allows you to do stuff at the IdP after authentication is complete"

something like

"Auth Proc Filters allows you to mutate of the attributes passed back from the IdP before Moodle handles them and does the profile field mappings"

README.md
```diff
@@ -276,6 +276,33 @@ Suggested attribute mappings:
...
+
+Steps to implement the hook:
+* Create a plugin that will implement the hook (e.g local_hookimplement)
+* Define the hook function 'local_hookimplement_extend_auth_saml2_proc' in plugin's lib.php
+* The function should return array of SimpleSaml Auth Proc Filters.
+
+Example:
+```php
+function local_hookimplement_extend_auth_saml2_proc {
+   // Do your data validation
```
brendanheywood: I'd remove this line

README.md
```diff
@@ -276,6 +276,33 @@ Suggested attribute mappings:
...
+
+   return [
+      51 => array(
+         'class' => 'core:AttributeMap',
+         'oid2name'
+      )
+   ]
```
brendanheywood: can you please add a more realistic example with the core:php

Issue #395, opened 2020-02-25
Unit test failure on latest version


ledship:
**What happened?**
Running on the latest T12 and Moodle 3.8 with the most recent version of auth_saml2 the locallib test fails on the data set "13. dual: p, param: null, session: false" ('passive', null, false, true)
Command for reference: 
`vendor/bin/phpunit --colors auth_saml2_locallib_testcase auth/saml2/tests/phpunit/locallib_test.php`

**What you expected:**
The unit test should of passed, this wasn't happening a few weeks ago so I assume a change in one of the recent commits has caused this.

ledship:
Looks like it's caused by https://github.com/catalyst/moodle-auth_saml2/commit/8d53164f69f9f435b58fd880094a741424412f0d

Issue #372, opened 2019-11-06
ispassive bug when falling back to manual login


brendanheywood:
**What happened?**

1) setup is passive on idp / sp
2) login but don't login at the idp, should to the redirect bounce and take you to the moodle login page
3) Enter a dud username and password
4) it tries that, fails, then redirects to saml and then back

**What you expected:**
Step 4 should not redirect to saml again. We should store the ispassive thing in both the query param in the errorUrl and also store it in the temp anon session

brendanheywood:
@kristian-94 didn't you fix this one?

PR #396, opened 2020-02-25
Fix issue #395

ledship:


ledship:
Looks like this breaks #372, will fix and push new commit.

PR #401, opened 2020-03-05
honor moodle proxy settings (issue #161)

jboulen:
Behind a http proxy, it should fix network issue with \auth_saml2\task\metadata_refresh task and when an administrator edits idp metadata settings. #161

### DIFF
--- a/classes/admin/setting_idpmetadata.php
+++ b/classes/admin/setting_idpmetadata.php
@@ -183,6 +183,10 @@ class setting_idpmetadata extends admin_setting_configtextarea {
      * @return idp_data[]
      */
     public function get_idps_data($value) {
+        global $CFG;
+
+        require_once($CFG->libdir.'/filelib.php');
+
         $parser = new idp_parser();
         $idps = $parser->parse($value);
 
@@ -192,7 +196,7 @@ class setting_idpmetadata extends admin_setting_configtextarea {
                 continue;
             }
 
-            $rawxml = @file_get_contents($idp->idpurl);
+            $rawxml = \download_file_content($idp->idpurl);
             if ($rawxml === false) {
                 throw new setting_idpmetadata_exception(
                     get_string('idpmetadata_badurl', 'auth_saml2', $idp->idpurl)
--- a/tests/phpunit/setting_idpmetadata_test.php
+++ b/tests/phpunit/setting_idpmetadata_test.php
@@ -134,6 +134,7 @@ class setting_idpmetadata_test extends advanced_testcase {
 
     public function test_it_returns_error_if_metadata_url_is_not_valid() {
         $error = self::$config->validate('http://invalid.url.metadata.test');
+        self::assertDebuggingCalled();
         self::assertContains('Invalid metadata', $error);
         self::assertContains('http://invalid.url.metadata.test', $error);
     }


classes/metadata_fetcher.php
```diff
@@ -70,6 +70,30 @@ public function fetch($url, $curl = null) {
             'CURLOPT_RETURNTRANSFER' => true,
             'CURLOPT_NOBODY'         => false,
         ];
+
+        // Set proxy.
+        if (!empty($CFG->proxyhost) && !is_proxybypass($url)) {
```
brendanheywood: shouldn't this whole section be redundant and handled already by the core curl() class?

jboulen: I don't think that core curl() class has an option to automatically bypass http proxy for some domains.

I based my code on how Moodle handle http proxy with [native authentification plugin CAS](https://github.com/moodle/moodle/blob/73f8c56dfca5dd4028845f1fbbaac0a83e43caf9/auth/cas/auth.php#L195). There is also another example on [mnet lib](https://github.com/moodle/moodle/blob/73f8c56dfca5dd4028845f1fbbaac0a83e43caf9/mnet/lib.php#L74).

brendanheywood: There is something I'm missing here.

Out of the box the curl() library handle all proxies, if set completely automatically across the board. The code in cas and mnet looks like junk to me honestly and should be fixed in core.

You can use the curl class and force the proxy on like this:

$c = new curl(array('proxy'=>true));

but if you leave it as null then it auto detects is based on the url being external or not, and based on the whitelists in $CFG->proxybypass, all this logic is here inside the curl lib:

https://github.com/moodle/moodle/blob/master/lib/filelib.php#L3070-L3097

I just tested a small standalone script and it works. You are probably already aware of this because in your diff in the file above to use download_file_content() which uses curl() under the hood so we get this totally for free.

Buuttt.... as far as I can tell this metadata_fetcher.php  is completely dead code!?? When you save the metadata in the gui, then it just calls your download_file_content() code, and when it is refreshed via cron, it also just re-uses the admin setting class:

https://github.com/catalyst/moodle-auth_saml2/blame/master/classes/task/metadata_refresh.php#L96-L100

I think class this was refactored into irrelevance 2-3 years ago and should have been removed completely back then

So I think this patch is as simple as the 3 lines in the first file. I do think however that you should add some timeouts like:

$rawxml = download_file_content($idp->idpurl, null, null, false, 20, 10);

jboulen: You are totally right ! On each request with curl class, [this function](https://github.com/moodle/moodle/blob/73f8c56dfca5dd4028845f1fbbaac0a83e43caf9/lib/filelib.php#L3331) is applied to know if proxy settings should be bypassed or not.

About `auth_saml2\metadata_fetcher` class, I let you remove this class in another commit ?

I will rebase my work, and apply timeout values as suggested.

Thank you for your feedback !

brendanheywood: yup cool I'm remove the other old stuff thanks heaps

jboulen: I removed changes on `auth_saml2\metadata_fetcher` class. It should be good now.

Issue #381, opened 2019-11-18
auth_saml2_idps table schema alignment issue, columns have the incorrect default value


kristian-94:
If installing an older version of the plugin and upgrading, the upgrade will run this script: 
https://github.com/catalyst/moodle-auth_saml2/blob/master/db/upgrade.php#L221

This sets the default to be null, when in the install.xml, it is '0' https://github.com/catalyst/moodle-auth_saml2/blob/master/db/install.xml#L28

This gives us the following output from check_database_schema.php:
```
php admin/cli/check_database_schema.php 
-------------------------------------------------------------------------------
auth_saml2_idps
 * column 'activeidp' has default 'NULL', expected '0' (I)
 * column 'defaultidp' has default 'NULL', expected '0' (I)
 * column 'adminidp' has default 'NULL', expected '0' (I)
```
The default should probably be '0', since that is what moodle inserts. That means the install.xml is correct and this needs to be fixed up in the upgrade script. We need to add a new upgrade that fixes this for existing saml2 installs.

kristian-94:
Fixed now

PR #403, opened 2020-03-18
Fix upgrade scripts: columns in auth_saml2_idps should have default '…

kristian-94:
For #381


brendanheywood:
This one looks good to me, I updated the branch good to merge when green again

PR #404, opened 2020-03-18
Trigger user_update which other plugins could use

dlnsk:
I think we should trigger user_update event because we update user data. It can be important for other plugins which use user's data/
I'm using https://github.com/catalyst/moodle-local_cohortauto to put users in cohorts automatically and it must change cohorts immediately after update.
We should update user after profile because profile it is a part of user's data and it is very strange to trigger event before profile update.


brendanheywood:
thanks @dlnsk that seems like some missed from the beginning of time

Issue #407, opened 2020-04-01
More logging of cert generation and refreshing


brendanheywood:
**What you expected:**

* [ ] When the cert is deleted log something to error log
* [ ] when the cert is generated, or regenerated, log to error log

* [x] could also emit a moodle log event and / or config log event, more the merrier

kristian-94:
@brendanheywood  do you want to check this one again now

PR #413, opened 2020-04-09
Add cert_regenerated event to log when saml certs are regenerated #407

kristian-94:
#407  
    - Add unit test for regenerate form submission
    - Add unit tests for event being emitted on regeneration
    - Refactor form processing to locallib function


brendanheywood:
Looks pretty good, what is up with travis?

kristian-94:
@brendanheywood  because of this tracker: https://tracker.moodle.org/browse/MDL-62907

The logs are stored as json now but serialized in moodle < 3.7. My code does json_decode($logrecord->other) to check if the logs have been generated properly. 

I am thinking we might need to make a code check for moodle version and do unserialize vs json_decode. Alternatively there might be something we can do with regex or just strpos.

tests/phpunit/form_regenerate_test.php
```diff
@@ -0,0 +1,95 @@
...
+        self::assertSame('moodle', $formdata->stateorprovincename);
+        self::assertSame('vetmoodle', $formdata->organizationname);
+        self::assertSame('moodle', $formdata->organizationalunitname);
+        self::assertSame('moodle', $formdata->commonname);
+        self::assertSame('noreply@moodle.test', $formdata->email);
+        self::assertSame(3650, $formdata->expirydays);
+        self::assertSame('Regenerate', $formdata->submitbutton);
+
+        // Get log records and check for presence of the cert_regenerated event.
+        $records = $DB->get_records_select('logstore_standard_log', "eventname = ?", ['\auth_saml2\event\cert_regenerated']);
```
brendanheywood, line 69: This is wrong, never reach into the logs directly, use the log reader api

https://docs.moodle.org/dev/Logging_2#Reading_API

brendanheywood:
That feels totally wrong. You should not be reaching into the logs internals, it should have an api and you get clean data out the other side and don't care how it is stored.

This code here is wrong:

https://github.com/catalyst/moodle-auth_saml2/pull/413/files#diff-61394462ae6c02e5b4661df99a6e7947R69

kristian-94:
@brendanheywood , thank you for that, see how that looks to you now. We have it passing travis

classes/event/cert_regenerated.php
```diff
@@ -0,0 +1,57 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+/**
+ * cert_regenerated class.
+ *
```
brendanheywood, line 19: where is your name? :)

locallib.php
```diff
@@ -475,5 +477,45 @@ function auth_saml2_get_default_idp() {
...
+        'organizationalUnitName' => $fromform->organizationalunitname,
+    );
+    $numberofdays = $fromform->expirydays;
+
+    $saml2auth = new \auth_plugin_saml2();
+    $eventdata = [
+        'reason' => "regenerated in saml settings page",
+        'userid' => $USER->id,
+    ];
+    cert_regenerated::create(['other' => $eventdata])->trigger();
```
brendanheywood, line 505: should we swap this order and only emit the event if the cert creation was a success?

kristian-94: I suppose since an error would be logged anyway, we don't need the event to be emitted as well in that case. I have made that change now

classes/event/cert_regenerated.php
kristian-94: i've put that in

locallib.php
```diff
@@ -475,5 +477,49 @@ function auth_saml2_get_default_idp() {
...
+        ];
+        cert_regenerated::create(['other' => $eventdata])->trigger();
+    }
+
+    // Also refresh the SP metadata as well.
+    $file = $saml2auth->get_file_sp_metadata_file();
+    @unlink($file);
+
+    // Can't redirect in a unit test.
+    if (empty($error) && !PHPUNIT_TEST) {
```
brendanheywood, line 517: If code is aware of unit tests then usually thats a code smell of the interface not being right.

A better solution would be to redirect in the calling code, if and only if the result was null. And then you can cleanly unit tests without this hack

kristian-94: @brendanheywood  I have moved that redirect now

classes/event/cert_regenerated.php
```diff
@@ -0,0 +1,58 @@
...
+ */
+class cert_regenerated extends base {
+    /**
+     * Returns non-localised event description with id's for admin use only.
+     *
+     * @return string
+     */
+    public function get_description() {
+        $byuser = '';
+        $this->other['userid'] ? $byuser .= "by user '{$this->other['userid']}'" : null;
```
brendanheywood, line 46: I think a simple if would be more readable here. also '' could default to 'unknown'

regenerate.php
```diff
@@ -22,6 +22,8 @@
  * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
  */
 
+use auth_saml2\event\cert_regenerated;
```
brendanheywood, line 25: is this used? can't see it in the diff

kristian-94: You're right, I've removed this now

classes/event/cert_regenerated.php
kristian-94: Have updated it to read more clearly now

tests/phpunit/form_regenerate_test.php
kristian-94: Have done this now, we're using log manager like 
```
$logmanger = get_log_manager();
        $readers = $logmanger->get_readers('\core\log\reader');
        $reader = reset($readers);
        $eventarray = $reader->get_events_select("eventname = ?", ['\auth_saml2\event\cert_regenerated'], 'id ASC', 0, 0);

kristian-94:
@brendanheywood , this is passing travis now, and your suggestions should all be implemented too

PR #418, opened 2020-04-21
Restrict automatic account creation to users in "allowed email domains"

### LINKED ISSUE #412: Honor the $CFG->allowemailaddresses setting
Hello

We have an IdP that authenticates users from multiple domains.  I have a requirement that users who are in a particular domain (let's say example.edu) should have Moodle accounts automatically created, but users whose accounts belong to any other domain should not.

But we still want to allow some users who are not part of example.edu to log in to Moodle.  We would do this by creating a Moodle account for them.

There is related logic here

https://github.com/catalyst/moodle-auth_saml2/blob/7241eaa54e9078df4ef1b475e5b7a19e55b2bfaa/auth.php#L586

Would you be interested in a PR that allows us to set "allowed domains" for automatic provisioning, but does not block users who already have a Moodle account from logging in?

porcospino:
Moodle has a core setting -- "allowemailaddresses" -- that you can use
to restrict access to email accounts in particular domains. It is used
by the "Email-based self-registration" and "OAuth2" authentication
plugins.

This change allows automatic account creation for users whose mail
attribute matches "allowemailaddresses".  But it does not prevent
you from creating a Moodle account for a user who has an address in
a different domain, and is authenticated by your IdP.

Fixes #412


porcospino:
Travis is giving this PR a hard time for no good reason.  Don't make me push a whitespace change! :-D

danmarsden:
Thanks Chris - kicked them both manually - hopefully that will help! - patch looks good to me, hopefully someone here will get a chance to test and merge it - thanks!

brendanheywood:
This patch is good, I've just tweaked the comment and updated the branch and will merge when green

PR #419, opened 2020-04-28
Add support for a mock IdP for use in Behat tests

sammarshallou:
Adds a minimal mock IdP and simple Behat steps which can be used to easily test aspects of this plugin, or to accurately simulate a SAML login as part of tests elsewhere in a system.

This makes accurate Behat tests super easy. Here is an example scenario (this is the complete scenario, no background needed) to test passive login:

```
  Scenario: Passive login (logged in)
    Given the following "users" exist:
      | username | auth  | firstname | lastname |
      | student1 | saml2 | Eigh      | Person   |
    And the mock SAML IdP is configured
    And the following config values are set as admin:
      | auth      | saml2 |            |
      | duallogin | 2     | auth_saml2 |
    When I follow "Log in"
    And the mock SAML IdP allows passive login with the following attributes:
      | uid        | student1 |
    Then I should see "Course overview"
    And I should see "Eigh Person" in the ".userbutton" "css_element"
```

Travis passes after this change, and as it only affects files within the 'test' folder, I think it should be quite a safe one to include.

By the way the reasons I want this are:
a) I'm planning to submit a second enhancement shortly but I want to be able to automatically test it so at least **I** believe it works. :)
b) We have a custom local plugin that does things based on the SAML attributes, and although it's possible to simulate this without actually using a SAML login, it would be more realistic to have tests that use the whole process. So I'd like to use these test steps in my custom plugin too.


sammarshallou:
Note: I do see there is already Behat support for a fake provider using a fake IP address (which is necessary for SimpleSAMLphp's IdP to work as it requires a different host), but this requires special treatment (hosts file etc.) to work so it's hard to set up, and it doesn't look as easy to use in tests, so I think this one is better for most uses. Obviously that's just my opinion...

brendanheywood:
> I do see there is already Behat support for a fake provider using a fake IP address

Yup, I'm leaning toward ripping the old stuff out completely.

sammarshallou:
I've made an updated version which replaces the existing code and updates all the existing tests (where necessary). All the Behat tests now pass on my machine. 

But it doesn't run in Travis because the recent node version change doesn't play well with moodle-plugin-ci; this is annoying (and hard to completely fix, I've been looking at it).

sammarshallou:
Right, this fixed branch - which temporarily includes the commit I did for the other pull request https://github.com/catalyst/moodle-auth_saml2/pull/427 - now passes Travis. I had to make another change so the fixture works OK on Firefox 47... I think it's OK though.

dmitriim:
Hi @sammarshallou 
Thanks for working on this one. It looks very good and I'm really keen to get it in.
Will you be able to rebase your changes on top of master?

Issue #408, opened 2020-04-01
Make certificate locking persistent and not based on files alone


brendanheywood:
**What happened?**

If you lock the certificates all this does is chmod the files. If the files themselves are gone, or if only 1 is gone they still regenerate even if they were locked.

So proposing to also store the locked state in config, and if the files should be locked then never regenerate them, even if they don't exist.

**What you expected:**

PR #420, opened 2020-04-30
Locked and Unlocked generate cert unit test #408

guillogo:
Add: 
- Unit Test -> if locked not generate the cert, if unlocked generate the cert
- Change the permissions in order to regenerate if unlocked
- a Lang String Message


kristian-94:
Closing, this is now part of #421  instead

PR #421, opened 2020-04-30
Store the certs locked state in config #408

kristian-94:
- Don't regenerate certs if they are locked
- Unit test for locked/unlocked cert generation

#408


kristian-94:
Do we want to be able to unlock the files again from the config? (ie. chmod them back to being writable)

brendanheywood:
There was a use case for these being locked down on the server to the point the plugin couldn't overwrite them. I'd like to see that tested, and if the file cannot be chmod'ed back to readable then expose that error, and also don't unset the locked config item

classes/form/lockcertificate.php
```diff
@@ -53,6 +53,19 @@ protected function definition() {
         $html .= html_writer::end_div();
 
         $mform->addElement('html', $html);
+
+        $html  = html_writer::start_div('warning');
```
brendanheywood, line 57: .warning isn't a css class, curious why it is here at all?

classes/form/lockcertificate.php
```diff
@@ -53,6 +53,19 @@ protected function definition() {
...
+
+        if (get_config('auth_saml2', 'certs_locked') == false) {
+            $html .= $OUTPUT->notification(get_string('certificatelock_unlockedmessage', 'auth_saml2'), 'warning');
+        } else {
+            $html .= $OUTPUT->notification(get_string('certificatelock_lockedmessage', 'auth_saml2'), 'warning');
+        }
+        $html .= html_writer::end_div();
+
+        $mform->addElement('html', $html);
+        $mform->addElement('submit', 'unlockcertsbutton', get_string('certificatelock_unlock', 'auth_saml2'));
```
brendanheywood, line 67: move this to a button group instead of add_action_buttons

certificatelock.php
```diff
@@ -43,8 +43,19 @@
     if ($form->is_submitted()) {
 
         $certfiles = array($saml2auth->certpem, $saml2auth->certcrt);
-        foreach ($certfiles as $certfile) {
-            chmod($certfile, 0440);
+        if (isset($data->unlockcertsbutton)) {
+            // Change the permissions in order to regenerate if unlocked.
+            foreach ($certfiles as $certfile) {
+                chmod($certfile, 0660);
```
brendanheywood, line 49: there is a moodle setting which I think we should use for this

$CFG->filepermissions

certificatelock.php
```diff
@@ -43,8 +43,19 @@
...
+            }
+            // Store the unlocked state in config.
+            set_config('certs_locked', '0', 'auth_saml2');
+        } else {
+            foreach ($certfiles as $certfile) {
+                chmod($certfile, 0440);
```
brendanheywood, line 55: the reverse it true too, set this to $CFG->filepermissions bit masked with 0440

classes/form/lockcertificate.php
kristian-94: @brendanheywood  this is done in a couple other places that I can see, 
https://github.com/catalyst/moodle-tool_lockstats/blob/master/classes/form/reset.php#L55
https://github.com/catalyst/moodle-tool_crawler/blob/master/classes/form/reset.php#L62

As well as in this file for the other warning message

kristian-94: It makes the text bold

kristian-94:
@brendanheywood  this one should be ok now if you want to check it out

regenerate.php
```diff
@@ -40,11 +40,12 @@
 $error = '';
 
 if ($fromform = $mform->get_data()) {
-    $error = auth_saml2_process_regenerate_form($fromform);
-    if ($error === '') {
-        redirect("$CFG->wwwroot/admin/settings.php?section=authsettingsaml2");
+    // Check if the certificates are locked, in that case don't regenerate.
```
brendanheywood, line 43: I think this checking logic isn't in the best place, I think create_certificate is where the most important check should be, and if it fails it throws a hard exception. Then its up to whatever that calls it to handle it, or not, and avoid it in advance



auth_saml2_process_regenerate_form

kristian-94: @brendanheywood  I moved the check into the `create_certificates` function now. The form does a try catch so that we see the error in the ` $OUTPUT->error_text` area which I think looks nice.

We unit test that if we call  `create_certificates` directly while the certs are locked, we throw a saml2 exception.

PR #424, opened 2020-05-06
SAML2: Fix customizable icon

fabianbatioja:
When $idp->logo has the URL of the logo, 2 notices appear on the login page. This happens because Moodle Core expects the idplist icon item to be an instance of pix_icon. This seems to be related to issue79


brendanheywood:
thanks @fabianbatioja

PR #426, opened 2020-05-12
Auto login / isPassive support improvements

sammarshallou:
**Note**: This pull request currently includes 2 commits but the first commit is just my other pull request about making Behat testing easier https://github.com/catalyst/moodle-auth_saml2/pull/419. 

This feature is a new configuration option that allows users to configure auth_saml2 to attempt passive login automatically, without the user having to go to a login page.

This is useful for servers that allow users to access without being logged in (e.g. public access), but where some users might be already logged on to the institutional identity provider.

There are two options: try at the start of a session, or try every time a cookie changes. The first option is simple but means it won't detect login if you go to the server before logging in. The second option can be useful if you are able to configure your identity provider to set a cookie, within the domain used by the Moodle system, when the user logs in.

In both cases, it will not do anything when you are already logged in (this is an autologin feature, not autologout).

### DIFF
--- a/classes/auto_login.php
+++ b/classes/auto_login.php
@@ -1,0 +1,226 @@
+<?php
+// This file is part of Moodle - http://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+/**
+ * Auto-login class.
+ *
+ * @package   auth_saml2
+ * @copyright 2020 The Open University
+ * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+
+namespace auth_saml2;
+
+use auth_saml2\admin\saml2_settings;
+
+defined('MOODLE_INTERNAL') || die();
+
+/**
+ * Auto-login class.
+ */
+class auto_login {
+    /** @var bool True if already considered auto-login on this page. */
+    protected static $processed;
+
+    /**
+     * Called on every page to give an opportunity to auto-login.
+     *
+     * This is called on most pages via the after_require_login callback, and also by the
+     * before_http_headers callback, so it might be called twice. We only want to run it once.
+     *
+     * The auto-login system is only useful on pages which do not already require a proper login.
+     * (For example the user might be trying to access the site homepage, or a course page which
+     * allows guest access.) In this case, the auto-login system lets us attempt to log in if
+     * we think the user might be logged into the SAML identity provider, so that they automatically
+     * get access to the extra controls that may appear on that page if they are logged in to their
+     * own account.
+     */
+    public static function process() {
+        global $SESSION, $SCRIPT;
+
+        if (self::$processed) {
+            return;
+        }
+        self::$processed = true;
+
+        // Do not apply this code in auth or login scripts, because those are likely to be already
+        // handling login (either auto-login or not). Users do not need to start auto-login from
+        // these pages, only from 'normal' Moodle pages they are trying to access.
+        if (preg_match('~^(/auth/|/login/)~', $SCRIPT)) {
+            return;
+        }
+
+        // If this is not a GET request, don't try autologin.
+        if ($_SERVER['REQUEST_METHOD'] !== 'GET') {
+            return;
+        }
+
+        // Only do anything if the user isn't already logged in, and the auth plugin is enabled.
+        if (!self::check_not_logged_in_and_enabled()) {
+            return;
+        }
+
+        // Get config.
+        $auth = get_auth_plugin('saml2');
+        switch ($auth->config->autologin) {
+            // Turned off.
+            case saml2_settings::OPTION_AUTO_LOGIN_NO:
+                return;
+
+            // Login once per session. Don't try login if we already tried this session.
+            case saml2_settings::OPTION_AUTO_LOGIN_SESSION:
+                if (!empty($SESSION->auth_saml2_triedautologin)) {
+                    return;
+                }
+                break;
+
+            // Login when a cookie exists or changes. Don't try login if we already tried for this
+            // cookie value, or the cookie isn't set.
+            case saml2_settings::OPTION_AUTO_LOGIN_COOKIE:
+                // If the cookie setting isn't filled in, do nothing.
+                if (empty($auth->config->autologincookie)) {
+                    return;
+                }
+                // If the cookie isn't set, do nothing.
+                if (empty($_COOKIE[$auth->config->autologincookie])) {
+                    return;
+                }
+                // If the cookie is set to the same value it was last time we looked, do nothing.
+                $currentcookie = $_COOKIE[$auth->config->autologincookie];
+                if (!empty($SESSION->auth_saml2_lastautologincookie) &&
+                        $SESSION->auth_saml2_lastautologincookie === $currentcookie) {
+                    return;
+                }
+                break;
+        }
+
+        // Check the plugin is configured (it probably is since they set up the autologin option).
+        // Note: This check is after the switch because it's highly likely that we will exit there.
+        // is_configured is not a costly function call but it does make some filesystem checks.
+        if (!$auth->is_configured()) {
+            return;
+        }
+
+        // Record in session that we attempted login.
+        switch ($auth->config->autologin) {
+            case saml2_settings::OPTION_AUTO_LOGIN_SESSION:
+                $SESSION->auth_saml2_triedautologin = true;
+                break;
+
+            case saml2_settings::OPTION_AUTO_LOGIN_COOKIE:
+                $SESSION->auth_saml2_lastautologincookie = $currentcookie;
+                break;
+        }
+        $SESSION->auth_saml2_autologinattempt = true;
+
+        // Now actually try to log in!
+        self::login($auth);
+    }
+
+    /**
+     * Confirms that the user is not logged in, and that auth_saml2 is enabled.
+     *
+     * @return bool True if the user is not logged in (or is guest), and saml2 is enabled
+     */
+    protected static function check_not_logged_in_and_enabled() {
+        // If they are already logged in, we don't need autologin.
+        if (isloggedin() && !isguestuser()) {
+            return false;
+        }
+
+        // Check if this plugin is enabled.
+        $enabled = get_enabled_auth_plugins();
+        return in_array('saml2', $enabled);
+    }
+
+    /**
+     * Tries to auto-login by making a passive login request.
+     *
+     * Also called once redirected back after a successful request.
+     *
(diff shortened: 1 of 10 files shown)


brendanheywood:
thanks heaps @sammarshallou 

I've only had a very superficial look:

1) At some point ages ago there was a real embedded idp for behat testing, I'm honestly not sure what its status is:

https://github.com/catalyst/moodle-auth_saml2/tree/master/tests/idp

If your patch does completely replace this can you please remove the old stuff please?

2) On the second commit,  isPassive was kinda half supported already so it would be good to double check if this is needed and if there is a gap align the two as needed (this is probably a missing documentation problem)

```
auth.php:535:        $passive = $this->config->duallogin == saml2_settings::OPTION_DUAL_LOGIN_PASSIVE;
auth.php-536-        $passive = (bool)optional_param('passive', $passive, PARAM_BOOL);
auth.php-537-        $params = ['isPassive' => $passive];
auth.php-538-        if ($passive) {
auth.php-539-            $params['ErrorURL'] = "{$CFG->wwwroot}/login/index.php?saml=0";
auth.php-540-        }
```

sammarshallou:
Thanks for taking a look Brendan, I appreciate it, I know everyone's busy!

Re first point - I'll figure out the existing embedded IdP for Behat in more detail and do a full replacement.

Re second point, about passive support - unless I've missed something, this doesn't duplicate the existing support. The difference is, the existing passive support only works on the login screen. For most Moodle installs (including some of ours) where login is usually required e.g. to access a course, this is fine as you will immediately get to a screen where you're prompted to login, so then it will do the passive login right away. The issue for us is that we have a couple of sites that are largely open to the public, so you can go to many pages without logging in, which means it doesn't do the passive login - and e.g. staff who are already logged in to the IdP don't get the various edit links, because they're still just guest on moodle.

classes/auto_login.php
```diff
@@ -0,0 +1,212 @@
...
+     */
+    public static function consider() {
+        global $SESSION, $SCRIPT;
+
+        if (self::$considered) {
+            return;
+        }
+        self::$considered = true;
+
+        // Don't do this on any auth or login scripts!
```
dmitriim, line 52: Can you please explain why not?

lang/en/auth_saml2.php
```diff
@@ -33,6 +33,12 @@
 $string['auth_saml2blockredirectdescription'] = 'Redirect or display message to SAML2 logins based on configured group restrictions';
 $string['autocreate'] = 'Auto create users';
 $string['autocreate_help'] = 'If users are in the IdP but not in moodle create a moodle account.';
+$string['autologin'] = 'Auto-login';
+$string['autologin_help'] = 'Where guest access is allowed, automatically log users into Moodle with a real user account if they are logged in to the IdP (using passive authentication).';
```
dmitriim, line 37: "Where guest access is allowed" this bit is a bit confusing. Guest access can mean different for different parts of moodle. What do you mean here?

classes/auto_login.php
```diff
@@ -0,0 +1,212 @@
...
+                if (!empty($SESSION->auth_saml2_lastautologincookie) &&
+                        $SESSION->auth_saml2_lastautologincookie === $currentcookie) {
+                    return;
+                }
+                break;
+        }
+
+        // Just as a final check (now we're definitely going to try login), is the plugin actually
+        // configured? It probably is since they set up the autologin option, but...
+        if (!$auth->is_configured()) {
```
dmitriim, line 109: Not a big deal really, but for me it looks like this check should go first before we even go to switch options,.

classes/auto_login.php
```diff
@@ -0,0 +1,212 @@
...
+            return;
+        }
+
+        // If they are already logged in, we don't need autologin.
+        if (isloggedin() && !isguestuser()) {
+            return;
+        }
+
+        // Check if this plugin is enabled.
+        $enabled = \core_plugin_manager::instance()->get_enabled_plugins('auth');
```
dmitriim, line 68: There is a core get_enabled_auth_plugins. What are benefits to check if it's enabled this way?

classes/auto_login.php
```diff
@@ -0,0 +1,212 @@
...
+     * @param \moodle_url $url URL for redirecting to
+     */
+    public static function finish($success, \moodle_url $url) {
+        global $SESSION;
+        if (empty($SESSION->auth_saml2_autologinattempt)) {
+            return;
+        }
+        unset($SESSION->auth_saml2_autologinattempt);
+
+        // If they are already logged in, we don't need autologin.
```
dmitriim, line 181: seems like lines 181-201 are more or less the same in self::consider method. Should we abstract those lines to reduce code duplication?

classes/auto_login.php
```diff
@@ -0,0 +1,212 @@
...
+    /** @var bool True if already considered auto-login on this page. */
+    protected static $considered;
+
+    /**
+     * Called on every page to give an opportunity to auto-login.
+     *
+     * This is called on most pages via the after_require_login callback, and also by the
+     * before_http_headers callback, so it might be called twice. We only want to run it once.
+     */
+    public static function consider() {
```
dmitriim, line 44: I know that one of the hardest things in software development is picking up names :) what stands behind this method name? For me more standard auto_login::process would fit better.

classes/auto_login.php
```diff
@@ -0,0 +1,212 @@
...
+        global $SESSION, $SCRIPT;
+
+        if (self::$considered) {
```
dmitriim, line 47: this method got quite a few if statements and I can't find test cases for each of them. Would be good to cover those in bahat or php unit. Like emulate POST request (when ready for autologin) and check that we actually don't auto login and etc

sammarshallou:
Thank you for the review! All good comments, I'll work on this and respond to each one soon.

classes/auto_login.php
sammarshallou: At least in my head :) it was called 'consider' because the function's job is to 'consider' whether or not to do auto-login. But 'process' is fine and it certainly sounds like the previous name wasn't obvious, so I'll just change it as you suggest.

classes/auto_login.php
sammarshallou: To be honest I wasn't aware of get_enabled_auth_plugins. The code is a bit shorter with get_enabled_plugins (because you can just look for the array key instead of having to loop through them), but performance might be marginally worse as it has to do a cache lookup, and I guess it would be more obvious to use the special-purpose function, so I'll change it to get_enabled_auth_plugins.

classes/auto_login.php
sammarshallou: It was after the switch for performance reasons - I was trying to be very careful about performance because this code runs on every page. is_configured only really does a few file_exists checks which is not exactly a big deal, but even so, I think it might as well stay later - I'll add a comment to explain why it is there.

classes/auto_login.php
sammarshallou: I've broken out two of the checks into a function, we could move the is_configured check in there as well if you think the performance thing isn't worth retaining.

lang/en/auth_saml2.php
sammarshallou: I've changed it to 'On pages that allow guest access without login' - is that clearer? Or we could remove the word 'guest' if that is confusing? Basically what I'm trying to say is, if a page calls require_login and doesn't allow guest access, then auto-login doesn't apply, it's only on the situation where the page allows access without login that autologin could happen.

classes/auto_login.php
sammarshallou: One answer is that it breaks. :) But I've added a proper answer in code comment. I also added some more explanation to the phpdoc comment at the top of the function. Basically if you are already doing something to do with login, then that would not be the point to trigger auto-login. We only need to start auto-login process if the user is trying to view a normal page (that allows access without login, but where they might have better access e.g. edit buttons if they are logged in).

classes/auto_login.php
sammarshallou: Good point. I've added a new Behat scenario that tests most of the possible reasons including POST request, if the plugin is disabled, if the option is turned off.

sammarshallou:
I've now updated based on all the review comments (I think). Feel free to say if there are more things need changing or you disagree with anything, I'm happy to change it again.

Since I did a force push, it's not obvious from history what I changed in response to the review, so I'm attaching here a patch file of the differences. 

[autologinchanges.txt](https://github.com/catalyst/moodle-auth_saml2/files/5118178/autologinchanges.txt)

lang/en/auth_saml2.php
```diff
@@ -33,6 +33,12 @@
 $string['auth_saml2blockredirectdescription'] = 'Redirect or display message to SAML2 logins based on configured group restrictions';
 $string['autocreate'] = 'Auto create users';
 $string['autocreate_help'] = 'If users are in the IdP but not in moodle create a moodle account.';
+$string['autologin'] = 'Auto-login';
+$string['autologin_help'] = 'On pages that allow guest access without login, automatically log users into Moodle with a real user account if they are logged in to the IdP (using passive authentication).';
```
dmitriim, line 37: Hi @sammarshallou ! Thank you for you changes. 
Sorry but I maybe miss something, but I'm still not 100% clear why we apply autologin on pages that do not require login at all and ignore others? I would think that this would be helpful for IDPs (e.g. like OKTA) that do not support autologin.

sammarshallou: Yes, this feature is intended to only run on pages that don't require login (either because they literally don't call require_login, or because they allow automatic guest login).

The reason for this is that if the page isn't one of these, i.e. if it DOES require login (& not allow guest autologin), then the user will be redirected to the login screen. And auth_saml2 already has an automatic login feature once you get to the login screen (set 'duallogin' to 'Passive mode' if you also allow other logins, or 'No' if you only allow login from the one SAML IdP). So I didn't want to overlap with the existing feature which already works well.

In terms of whether it works on different IdPs (I don't know about OKTA sorry), this new feature requires passive mode support from the IdP, the same as the existing automatic login from login screen, so I don't think there should be any difference in terms of support.

sammarshallou:
I've rebased this (all the CI tests fail for master and this branch due to the other issue I pushed, but I did run them in a test branch including that fix and it passes still). Any chance of getting it in?

dmitriim:
Sorry @sammarshallou. It somehow disappeared from my radar. Unfortunately, I won't be able to finish up this review. From  the first glance the patch looks good and does what it supposed to do. However, I think it can be somehow used for for IDPs that do not support passive mode (like OKTA). But that could be done in a separate request when someone will need it. 

I'll ping @brendanheywood or maybe @danmarsden to check if they have some time.

sammarshallou:
Thanks @dmitriim, I understand not having time!

By the way if it would help, my project manager said we might have a (small) budget available that could pay for Catalyst time to finish code reviewing this & get it in - I'm not sure if that is in any way feasible, maybe all the relevant people are busy for the near future anyway. But if it would help, let me know and I'll ask our project manager to get in touch.

danmarsden:
Thanks for the ping @dmitriim  - I've had a look through the code and I couldn't see anything that I thought had to be changed - the fact that this is turned off by default, doesn't seem to affect anyone that has it turned off, comes with some good behat tests, would make me happy to +1 this.

@dmitriim if your reccommendation is to merge this in, I'm happy to +1 and hit the button :-)

dmitriim:
@danmarsden  yeah. Go ahead. I'm wondering why CI is not running?

danmarsden:
it's something weird - I can see it's passing on Sam's own repo - I'm guessing it will run (and pass) when I merge it.

danmarsden:
@sammarshallou had another PR that fixed that so I've merged it in and that seems to have fixed it.

sammarshallou:
Thank you both. :)

PR #427, opened 2020-05-18
Update Travis to use latest Node version, now required by Moodle tools

sammarshallou:
Travis is all failing because of the old Node version. This is an attempt to fix it.

Note that 'nvm install' with no number doesn't work because there is
no .nvmrc from Moodle available at that point, so I just changed the
number to 14 which is the current required.

Also fixes a missing NL at EOF in a test file which causes a failure
too.

This makes Travis mostly pass, but MOODLE_36_STABLE is still failing, some kind of error recompiling SASS. Looking at this https://github.com/blackboard-open-source/moodle-plugin-ci/issues/115 it seems that MOODLE_36_STABLE is failing for other people too. Not sure what to do about that...


sammarshallou:
Here's an actual working version. By the way, I tried avoiding the environment variable and using the Unix 'test' command to conditionally select node version, but I wasn't able to get this to work.

brendanheywood:
thanks @sammarshallou

PR #435, opened 2020-07-08
Match simplesaml session.duration to Moodle sessiontimeout

andrewhancox:
The session duration for simplesamlphp was set to 8hrs so if you set the Moodle session timeout to 1 hour you were re-authenticated by auth_saml2 without having been sent to the IDP.
The use case for me was:
Client sets IDP timeout to 1hr, Moodle session timeout too 15min. After 1hr and 1min of inactivity the Moodle session is silently renewed without the user hitting the IDP and being forced to re-authenticate.

### DIFF
--- a/config/config.php
+++ b/config/config.php
@@ -59,7 +59,7 @@ $config = array(
     // TODO \core_user::get_support_user().
     'timezone' => class_exists('core_date') ? core_date::get_server_timezone() : null,
 
-    'session.duration'          => 60 * 60 * 8, // 8 hours. TODO same as moodle.
+    'session.duration'          => (int)$CFG->sessiontimeout,
     'session.datastore.timeout' => 60 * 60 * 4,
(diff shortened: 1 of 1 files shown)


brendanheywood:
thanks @andrewhancox a slight issue with inactivity vs total duration, so you could be using moodle and keep its session alive while the internal saml session has expired, but this is still a step in the right direction

andrewhancox:
Totally agree but I needed a quick fix...
That said, I'm struggling to see a scenario where you would want the session in the auth plugin to be valid when the Moodle session has timed out. So far as I can see the only timeouts you would ever want are those set by the Moodle session and the IDP so this may as well be set to zero. Let me know if you come up with one...

brendanheywood:
My ideal architecture would be to remove the middle internal SSP session completely. I vaguely remember there being some edge cases, possibly to do with single signout needing enough of a valid session to do back channel sign sign out

PR #436, opened 2020-07-14
Redirect guest continue button for self enrolment

durzo:
Setup:

auth_saml2 : duallogin = no

http://moodle.local/enrol/index.php?id=2

If you have Guest access enabled and a course with allow self enrolment, Moodle displays a button that says "Continue" which performs a POST with no variables to /login/index.php

If the site is using auth_saml2 then the user gets stuck on the /login/index.php page with no way to login via saml.

This PR catches the empty post and redirects to the saml login page, fixing the guest self enrolment continue button.

### DIFF
--- a/classes/auth.php
+++ b/classes/auth.php
@@ -410,10 +410,16 @@ class auth extends \auth_plugin_base {
             $saml = 0;
         }
 
-        // Never redirect on POST.
+        // Sometimes redirect on POST to catch guest "continue" button in self enrolment.
         if (isset($_SERVER['REQUEST_METHOD']) && ($_SERVER['REQUEST_METHOD'] == 'POST')) {
-            $this->log(__FUNCTION__ . ' skipping due to method=post');
-            return false;
+            if (empty($_POST['username']) && empty($_POST['password'])) {
+                $this->log(__FUNCTION__ . ' redirecting due to missing username and password in POST (guest)');
+                $loginurl = new moodle_url('/auth/saml2/login.php');
+                redirect($loginurl);
+            } else {
+                $this->log(__FUNCTION__ . ' skipping due to method=post');
+                return false;
+            }
         }
 
         // Never redirect if requested so.


durzo:
I'm not sure how to fix the unit test here, sorry

brendanheywood:
thanks, looks pretty good. To fix the tests you'll need to cover that scenario here and add some test data to the dataprovider:

https://github.com/catalyst/moodle-auth_saml2/blob/master/tests/phpunit/locallib_test.php#L108-L159

durzo:
So I have tried a few different times and come up blank.

You can't test Moodle's redirect() because it throws an exception if CLI_SCRIPT is defined.

I modified test 17 to send a dummy username + password combo but now some other unit test is breaking with a 503 which I can't figure out.

tests/phpunit/locallib_test.php
```diff
@@ -111,6 +111,8 @@ public function test_should_login_redirect($duallogin, $param, $session, $expect
                 $_GET['SimpleSAML_Auth_State_exceptionId'] = '...';
             } else if ($param == 'post') {
                 $_SERVER['REQUEST_METHOD'] = 'POST';
+                $_POST['username'] = 'testuser';
```
brendanheywood, line 114: test_should_login_redirect gets call many times with different data, this is what the @dataProvider should_login_redirect_testcases is for. 

So instead of hard coding this, make $post a new function argument and then pass it in as a new test scenario in the data provider and default it to null in all the existing dataproviders

Admittedly this test function has way too much logic in it and needs a refactor, it should do almost nothing other that creating the preconditions for should_login_redirect

brendanheywood:
> You can't test Moodle's redirect() because it throws an exception if CLI_SCRIPT is defined.

Ah yes, thats exactly why we made this function in the first place. This function should not redirect (I just noticed another place it does too which is a bug) - it should only ever return something to somewhere else which then does the redirect. So the way foward is to change the contract of that function so that it only ever returns:

true = redirect to saml idp
false = do not redirect
string = redirect to the url in the string instead of the idp

durzo:
there is already a redirect() in that function so mine isn't the first - but there is no unittest for that one :(

brendanheywood:
Yeah if you can fix up your patch, I'll sort the existing one out

tests/phpunit/locallib_test.php
brendanheywood: just a heads up this existing dodgy unit test will get fixed in another PR which is being worked on now, eta a day or so, will make this a bit simpler

brendanheywood:
@durzo sorry for the long delay. I'm trying to reproduce this and fix up the unit tests but I'm really struggling with the preconditions to reproduce it. Can you outline what settings and what urls to hit to reproduce this please?

durzo:
enable "self enrolment" from /admin/settings.php?section=manageenrols
create a test course (e.g course id 5) and enable "self enrolment" from /enrol/instances.php?id=5
login as guest and access the course /course/view.php?id=5
you should see:
Screenshot from 2020-11-04 10-35-29
The Continue button is a POST form which is not intercepted by auth_saml2

brendanheywood:
I still can't reproduce this. Your steps above is what I'd already done, I go to my test course:

http://moodle.local/enrol/index.php?id=2

I'm not logged in so it punts me to the login page, I click login as guest and redirects me back to the course and I get that 
image

Then I click Continue and it takes me back to the login page again. I can either login manually or use saml and both work fine. If I turn off dual auth then it takes me to the idp and I can still login fine and then back to the course again.

I have also flipped $CFG->autologinguests and it all still works for me. This is on bleeding edge moodle and latest saml

durzo:
On our site, clicking Continue brings you to the login page which gives no option to use saml - only manual.
This is Moodle 3.8.3 with auth_saml2 version 2019110702

Screenshot from 2020-11-05 11-02-54

I can upgrade the version of auth_saml2 if you think it will change, but we can't upgrade core Moodle until the next client upgrade window however I can clone the site and upgrade to 3.9 to try that.

brendanheywood:
That might be something unrelated and just config, hat is the conf->showidplink setting? This hides that link

https://github.com/catalyst/moodle-auth_saml2/blob/master/auth.php#L199

jtsafran:
Correct.  showidplink is hidden in this case as Disco is enabled and we want the login to look like Moodle without the IdP buttons.

The config you are missing to replicate the issue, I think, is auth_saml2 : duallogin = no.  That should redirect the user to the IdP (or in this case, the Disco service).

brendanheywood:
I had tried all permutations of dual login. The missing config is the disco service itself. I think the logic should be is simply if you have set a disco service then it always redirects. If the disco service comes back with no idp then we can cache that to avoid subsequent redirects. Would you agree with that logic?

jtsafran:
@brendanheywood my apologies.  I had missed this notification.

To be clear, by "the logic should be is simply if you have set a disco service then it always redirects" do you mean that the continue button always redirects?

jtsafran:
@brendanheywood wanted to ping you on this one and make sure you saw my last question.  If you mean that clicking continue will redirect you to the discourl, that would work great.  I don't think we want to force it before that as the course could have guest access enabled and not need a login.

Hopefully that makes sense!

brendanheywood:
hi @jtsafran / @durzo 

Just getting back to this old pr. The logic here might be broadly correct but can you please make it return a url as a string rather than redirecting directly itself. Please also add some data to the unit test dataProvider around this.

brendanheywood:
Also we have swapped our branching strategy to move to align with stables with breaking changes, so can you please reopen this against MOODLE_39_STABLE

tuanngocnguyen:
Hi @jtsafran @durzo 

This PR is closed due to removal of master branch.
Please follow Brendan's comment to create new PR

Regards

Issue #440, opened 2020-07-29
Database schema inconsistence


golenkovm:
Looks like after #320 was fixed some sites that have `auth_saml2` installed still have these columns with default value NULL. 
DB schema check throws:

```
auth_saml2_idps
• column 'activeidp' has default 'NULL', expected '0' (I)
• column 'defaultidp' has default 'NULL', expected '0' (I)
• column 'adminidp' has default 'NULL', expected '0' (I)
```

PR #441, opened 2020-07-29
Set correct default values for auth_saml2_idps table columns.

golenkovm:
This is a fix for #440

### DIFF
--- a/db/upgrade.php
+++ b/db/upgrade.php
@@ -306,5 +306,21 @@ function xmldb_auth_saml2_upgrade($oldversion) {
         upgrade_plugin_savepoint(true, 2020031800, 'auth', 'saml2');
     }
 
+    if ($oldversion < 2020072900) {
+        // Set default values to '0' (I) for activeidp, defaultidp and adminidp fields.
+        $table = new xmldb_table('auth_saml2_idps');
+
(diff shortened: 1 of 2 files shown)


db/upgrade.php
```diff
@@ -306,5 +306,21 @@ function xmldb_auth_saml2_upgrade($oldversion) {
         upgrade_plugin_savepoint(true, 2020031800, 'auth', 'saml2');
     }
 
+    if ($oldversion < 2020072900) {
+        // Set default values to '0' (I) for activeidp, defaultidp and adminidp fields.
+        $table = new xmldb_table('auth_saml2_idps');
```
brendanheywood, line 311: Is this a universal fix for everyone or does it need to be conditional if it detects the schema isn't right?

golenkovm: Is there a way to check the current default value for a given column?
I've tested this if the default value has been already set to 0 - `change_field_default()` works fine (changes nothing, but I guess that's fine).

golenkovm: I mean that this can be considered as a universal fix.
If the site has these columns with default `NULL` - columns will be altered to have default `0`,
If the site has these columns with default `0` - columns will be altered as well, but it won't cause any errors, default will be still `0`.

brendanheywood: ok perfect thats all I was checking

PR #449, opened 2020-08-19
Issue432 - fixing deep link with saml=on with after_config hook

kristian-94:
For #432 

Will add some readme docs

### DIFF
--- a/lib.php
+++ b/lib.php
@@ -1,0 +1,47 @@
+<?php
+// This file is part of Moodle - http://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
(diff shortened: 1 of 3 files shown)


lib.php
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * @package auth_saml2
+ * @license http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ * @copyright Catalyst IT
+ */
+
+/**
+ * Check if we have the saml=on param set. If so, disable guest access and force the user to log in with saml.
+ *
+ * @return boolean
```
brendanheywood, line 26: this function is void

lib.php
```diff
@@ -0,0 +1,41 @@
...
+ */
+function auth_saml2_after_config() {
```
brendanheywood, line 28: this hook is very raw and can be a little dangerous in edge cases just as during install or upgrade, run from cli or webservice etc. Can you wrap this in a generic try catch as this should never fail for any reason, but it can log debugging calls

kristian-94: @brendanheywood  I've wrapped it all in a try catch now

PR #456, opened 2020-09-17
Refactor regenerate screen to fix #455

### LINKED ISSUE #455: Notify users on certificate regeniration
Currently when you regenerate the certificate it displays error message only. Would be good to notify a user about successfully regenerated certificate. Otherwise it's a bit confused as you stay within the form without any ideas if what you did worked.

dmitriim:


mattporritt:
@dmitriim providing travis is happy, this is good to go.

dmitriim:
image

image

image

PR #471, opened 2020-10-02
Remove PHPMailer from auto loaded files

dmitriim:


mattporritt:
I'm not sure if it's the correct approach, maybe @brendanheywood has thoughts, but could we just remove phpmailer from the "parent" composer (simplesaml) file and then it's never brought in at all?

dmitriim:
@mattporritt @brendanheywood actually I can't see where  it comes from. It's been sitting in the generated simplesamlphp package that we throw. And then it gets autoladed with other vendor stuff. I guess this fix is the only one we can do.

dmitriim:
However, there is auth/saml2/.extlib/simplesamlphp/vendor/composer/installed.json I can try to remove it from there to check if that would fix ci tests

mattporritt:
@dmitriim give that a go, if that doesn't work, go with your first patch

dmitriim:
@mattporritt it doesn't work as seems like we don't run composer for auth_saml2, but autoload files are still getting used. So if @brendanheywood is ok, I'd merge it in.

PR #472, opened 2020-10-22
Allowing PHP Unit to work in version 8 and above

MartinSandberg:
Adding :void to stop PHP Unit tests in other plugins not fail because of missing :void in this


ScottVerbeek:
This would only be necessary in Moodle 310 and up. Created a new branch and combined these commits into one. See this new pull request https://github.com/catalyst/moodle-auth_saml2/pull/493 .

Issue #467, opened 2020-09-30
PHP lint fails


alexmorrisnz:
**What happened?**
PHP lint fails after the simplesamlphp library update yesterday in PR #465.  
Failing on php 7.2.24

```
PHP Parse error:  syntax error, unexpected end of file, expecting function (T_FUNCTION) or const (T_CONST) in server/auth/saml2/.extlib/simplesamlphp/vendor/symfony/config/Tests/Fixtures/ParseError.php on line 8\nxargs: /usr/bin/php7.2: exited with status 255; aborting
```

https://github.com/catalyst/moodle-auth_saml2/blob/master/.extlib/simplesamlphp/vendor/symfony/config/Tests/Fixtures/ParseError.php is the file causing the error

**What you expected:**
PHP lint to not fail

brendanheywood:
@Anupamatd  / @jwalits can you please make sure that when you merge things in that the linked issues are also closed

PR #475, opened 2020-10-30
Issue #467: Fix unexpected end of file in .extlib/simplesamlphp/vendo…

Anupamatd:
…r/symfony/config/Tests/Fixtures/ParseError.php


danmarsden:
this seems a bit weird to me.... - has this been reported upstream to symfony - I'd be more inclined to remove the tests than implement a code-change that probably breaks the symfony config tests. did you have a discussion with @brendanheywood on this?

Issue #484, opened 2020-11-25
Saml2 unit test problem - auth_saml2 leaves libxml errors in the error buffer causing other plugins to fail unit tests


kristian-94:
During saml2 unit tests, we call this part of the code: 
https://github.com/catalyst/moodle-auth_saml2/blob/master/classes/idp_parser.php#L72

This adds xml errors to an error buffer: https://www.php.net/manual/en/simplexml.examples-errors.php

We don't clear these after the saml2 testsuite runs.

We need to always clear errors, saml2 should finish unit tests and leave us in a clean state:
https://www.php.net/manual/en/function.libxml-clear-errors.php
`libxml_clear_errors()`

PR #485, opened 2020-11-25
Clear libxml error buffer after unit tests #484

kristian-94:
#484


brendanheywood:
Looks good to me merge when green, or greener than before

PR #493, opened 2021-01-05
Allowing PHP Unit to work in version 8 and above

ScottVerbeek:


ScottVerbeek:
The failing checks are all because of PHPDoc

danmarsden:
New 310stable branch looks incorrect for how this plugin manages branches. Latest code must always land in master branch first and then backported. If a patch breaks backwards compatibility then you would create a 3.9 and lower branch and master would be for 3.10 and higher. @brendanheywood might have further comments

danmarsden:
If a new branch is required you will also need to update the requires version in version.php and update the readme with details on the new branch and its supported versions

ScottVerbeek:
A new pull request, https://github.com/catalyst/moodle-auth_saml2/pull/497, has been opened.

PR #496, opened 2021-01-06
Master readme update

ScottVerbeek:


danmarsden:
This is looking even less normal... I'm away on leave but would be good to have a sanity check/convo on this before it lands when I get back next week? ... i think I know why this is being done but I'm not convinced its the right approach and I cant see any public comms on this... (naming stuff is hard)

ScottVerbeek:
Hi Dan, this was decided yesterday between @syn @brendanheywood @dmitriim and @ScottVerbeek in a meeting yesterday. We wanted a default branch convention. You'll see this returning in other plugins. We have documented this in out internal Wiki. Sorry about the confusion.

ScottVerbeek:
Closing this and the branch until this is resolved.

danmarsden:
@brendanheywood @dmitriim can you hold off.rolling this out until next week? I dont think this looks correct and we might have issues maintaining this convention moving forward with totara branches and I'd like to be consistent with our conventions across all plugins if we can.

PR #495, opened 2021-01-06
Updated README.md with new branches

ScottVerbeek:

PR #497, opened 2021-01-07
Fixes to Moodle 3.10 and later

ScottVerbeek:
These are pushed towards compatibility of Moodle 3.10. If this request gets accepted please also merge this https://github.com/catalyst/moodle-auth_saml2/pull/495 one because it'll update the README of that branch.


danmarsden:
I'm not convinced we need a new branch here - the other option is to drop PHP 7.0 support from the master branch - which probably isn't a bad thing considering it's been out of support from PHP for 2 years already.

danmarsden:
I've pushed through #508 as a replacement for this one which cherry-picks Scott's unit test fix and bumps the required PHP version to 7.2

PR #500, opened 2021-01-12
Prevent issues with t13 unit tests.

danmarsden:


danmarsden:
don't like this patch - wrote it quickly to get t13 working... setup/setuplib.php seem to be completely internal files and I can't see any use of accessing them dirrectly via http - setuplib.php definitely doesn't need a config include, and I suspect setup.php doesn't either so I'll test another option here.

PR #502, opened 2021-01-12
Fix deprecated calls to assertContains.

danmarsden:


danmarsden:
Note - this particular one breaks PHP 7.0 compatibilty, so we need to make a decision on which branches support php 7.0 before merging this.

danmarsden:
looks like this one breaks Moodle 3.5 compatibility - I'll take another look to see if we can prevent that before merging.

PR #504, opened 2021-01-12
Fix #503 - Convert to using expectException

### LINKED ISSUE #503: phpdoc expectedexception no longer supported
we need to convert these to use $this->expectException.

danmarsden:


danmarsden:
combining this one in issue501 PR.

PR #509, opened 2021-01-12
setup/setuplib.php are internal files - not called directly.

### LINKED ISSUE #489: Problems initialising unit tests on T13
**What happened?**

Totara 13 will print a stacktrace in `server/lib/init.php` when you include config.php multiple times, this is a problem with auth_saml2 because it loads `auth/saml2/settings.php` which calls off to the auth plugins function `get_ssp_version()` loading `setup.php` which loads config.php.

Line 105 in `server/lib/init.php` is the one printing the log.

**What you expected:**
There to not be a stacktrace printed into console when initialising phpunit.

danmarsden:
fixes #489 in a less funky way.

setuplib.php has no business including config.php, and setup.php seems to be only used by require('setup.php') calls within other scripts so as an internal file it shouldn't either.

This causes issues with Totara 13 (more detail in issue #489) but these files are not really designed to be called via http so should use the internal checks as included.

### DIFF
--- a/setup.php
+++ b/setup.php
@@ -22,11 +22,10 @@
  * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
  */
 
+defined('MOODLE_INTERNAL') || die();
+
 use auth_saml2\event\cert_regenerated;
 
-// @codingStandardsIgnoreStart
-require_once(__DIR__ . '/../../config.php');
-// @codingStandardsIgnoreEnd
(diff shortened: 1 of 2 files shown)


danmarsden:
btw - here's a link to the old patch that also fixes this for Totara but is a lot more funky and I don't like.
https://github.com/catalyst/moodle-auth_saml2/pull/500/commits/77f1d2fb4a6de97926ba480fa0e704df304d5363

brendanheywood:
This looks good to me

PR #512, opened 2021-01-14
Include moodle config file in metadata

alexmorrisnz:
https://github.com/catalyst/moodle-auth_saml2/blob/master/setup.php#L25 setup.php is now a moodle internal file,  https://github.com/catalyst/moodle-auth_saml2/blob/master/sp/metadata.php#L31 calls setup.php without including config.php causing setup.php to die.

### DIFF
--- a/sp/metadata.php
+++ b/sp/metadata.php
@@ -28,6 +28,7 @@
  * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
  */
 
+require_once('../../../config.php');
 require_once('../setup.php');
 require_once('../locallib.php');
 


sp/metadata.php
```diff
@@ -28,6 +28,7 @@
  * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
  */
 
+require_once('../../../config.php');
```
danmarsden, line 31: I think the "convention" these days is to use __DIR__
like:
require(__DIR__. '/../../config.php');

https://docs.moodle.org/dev/What_happens_when_you_require_config.php

PR #514, opened 2021-01-15
513 add password to redis

tholudwig:
From version 5.3 the password could already be transferred with redis->connect (…). But since not all instances use v5.3 or higher, the idea is to transfer the password later with redis->auth(...).
Another CFG variable auth_saml2_redis_password is introduced for this purpose. Authentication with a password is only possible if this is also set in config.php.

For more information see: https://github.com/phpredis/phpredis#connect-open

### DIFF
--- a/classes/redis_store.php
+++ b/classes/redis_store.php
@@ -110,6 +110,11 @@ class redis_store extends \SimpleSAML\Store {
         try {
             $redis = new \Redis();
             $redis->connect($CFG->auth_saml2_redis_server);
+
+            if (!empty($CFG->auth_saml2_redis_password)) {
+                $redis->auth($CFG->auth_saml2_redis_password);
+            }
+
         } catch (\RedisException $e) {
             throw new \coding_exception("RedisException caught with message: {$e->getMessage()}");
         }


danmarsden:
I'm inclined to close this as won't fix if redis 5.3 and higher support it already... especially as ubuntu 18.04 and higher seem to contain 5.4 and although ubuntu 16.04 has an older version, support for 16.04 ends this month.

I'm guessing there are some other distros that are still supported that use older versions of Redis - but do we need to support them in saml?

I'm also not a big fan of polluting the global $CFG var with plugin specific settings (I know we are already doing this, but it's not really a good idea to add more...)

@tholudwig - what distros are you trying to support that use an older version of Redis and might need this?

tholudwig:
Hi Dan,We are currently running on Ubuntu 18 and our fix just works fine for us.If you think there's no need for all other Moodle instances closed the issue is fine for me. But thanks for having a lookBest regars ThomasAm 09.04.2021 07:28 schrieb Dan Marsden ***@***.***>:
I'm inclined to close this as won't fix if redis 5.3 and higher support it already... especially as ubuntu 18.04 and higher seem to contain 5.4 and although ubuntu 16.04 has an older version, support for 16.04 ends this month.
I'm guessing there are some other distros that are still supported that use older versions of Redis - but do we need to support them in saml?
I'm also not a big fan of polluting the global $CFG var with plugin specific settings (I know we are already doing this, but it's not really a good idea to add more...)
@tholudwig - what distros are you trying to support that use an older version of Redis and might need this?

danmarsden:
maybe I'm missing something there? - your first post stated that your patch wasn't needed for redis 5.3 and higher, but you're using it on 18.04 which I think uses 5.4? - is it still needed with 5.4?

tholudwig:
Just to make things a bit clear. We are running Redis 6.x. On the first hand the patch was implemented cause our Redis server are password protected. During the patch we discovered to ways to get password protection working. For the patch we prepared for catalyst, we decided to go with the wider solution that works even for Redis < 5.3.

For me, it is just okay to skip the patch. Maybe I'll get the time to code the patch for a more modern Redis.


________________________________
Von: Dan Marsden ***@***.***>
Gesendet: Dienstag, 13. April 2021 01:45
An: catalyst/moodle-auth_saml2 ***@***.***>
Cc: Thomas Ludwig ***@***.***>; Mention ***@***.***>
Betreff: Re: [catalyst/moodle-auth_saml2] 513 add password to redis (#514)


maybe I'm missing something there? - your first post stated that your patch wasn't needed for redis 5.3 and higher, but you're using it on 18.04 which I think uses 5.4? - is it still needed with 5.4?

danmarsden:
Thanks @tholudwig - in that case I'll close this one off for now, but I think we could still be convinced to add this in future if it's useful for recent versions of redis.

thanks!

Issue #352, opened 2019-09-11
There is a few css and js files inside extlib which are not served with good caching headers


brendanheywood:
**What happened?**

* [x] these files are hitting nginx land, they should be cached at the varnish / cdn layer

**What you expected:**

* [ ] probably proxy these files and send version based cache headers

brendanheywood:
This is actually a bug, these files cannot be served from here at all:

https://github.com/catalyst/moodle-auth_saml2/blob/master/.extlib/simplesamlphp/templates/post.php#L7-L8

PR #525, opened 2021-02-23
Serve post files from web accessible location

alexmorrisnz:
For #352

### DIFF
--- a/.extlib/simplesamlphp/templates/post.php
+++ b/.extlib/simplesamlphp/templates/post.php
@@ -4,8 +4,8 @@
 <head>
     <meta http-equiv="content-type" content="text/html; charset=utf-8" />
     <title>POST data</title>
-    <script src="/<?php echo $this->data['baseurlpath']; ?>../.extlib/simplesamlphp/www/resources/post.js"></script>
-    <link type="text/css" rel="stylesheet" href="/<?php echo $this->data['baseurlpath']; ?>../.extlib/simplesamlphp/www/resources/post.css" />
+    <script src="/<?php echo $this->data['baseurlpath']; ?>resources/post.js"></script>
+    <link type="text/css" rel="stylesheet" href="/<?php echo $this->data['baseurlpath']; ?>resources/post.css" />
 </head>
 <body>
 
--- a/sp/resources/post.css
+++ b/sp/resources/post.css
@@ -1,0 +1,3 @@
+input#postLoginSubmitButton {
+    display: none;
+}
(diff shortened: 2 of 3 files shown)


danmarsden:
I haven't looked but is there a way to override the template without needing to change the upstream simplesaml lib? - it's easier if we can maintain a clean copy of upstreams simplesaml lib.

I'd also be tempted to check to see if the newer version of simplesamlphp (which we need to upgrade at some point anyway) - makes this any easier?

alexmorrisnz:
You could move the post.js/css files into `sp/`, then remove the changes made to the template reverting it back to upstream.
```
<script src="/<?php echo $this->data['baseurlpath']; ?>resources/post.js"></script>
    <link type="text/css" rel="stylesheet" href="/<?php echo $this->data['baseurlpath']; ?>resources/post.css" />
```
The files would then be served out of `/auth/saml2/sp/`

I think the newer version may make this harder as post.php is switched to the twig templating system.

.extlib/simplesamlphp/templates/post.php
```diff
@@ -4,8 +4,8 @@
 <head>
     <meta http-equiv="content-type" content="text/html; charset=utf-8" />
     <title>POST data</title>
-    <script src="/<?php echo $this->data['baseurlpath']; ?>../.extlib/simplesamlphp/www/resources/post.js"></script>
-    <link type="text/css" rel="stylesheet" href="/<?php echo $this->data['baseurlpath']; ?>../.extlib/simplesamlphp/www/resources/post.css" />
+    <script src="/<?php echo $this->data['baseurlpath']; ?>../post.js"></script>
+    <link type="text/css" rel="stylesheet" href="/<?php echo $this->data['baseurlpath']; ?>../post.css" />
```
brendanheywood, line 8: The original code is:
```
    <script src="/<?php echo $this->data['baseurlpath']; ?>resources/post.js"></script>
    <link type="text/css" rel="stylesheet" href="/<?php echo $this->data['baseurlpath']; ?>resources/post.css" />
```

so can we just revert back to that and put the files into here and then it should all be zero touch:

```
/auth/saml2/sp/resources/post.css
/auth/saml2/sp/resources/post.js
```

brendanheywood: and this should still work when we upgrade to the next version which dumps the php templates and moves to twig instead:

https://github.com/simplesamlphp/simplesamlphp/blob/master/templates/post.twig

brendanheywood: ah snap, didn't see the other comments, we're on the same page :)

brendanheywood:
This is good to go when green

alexmorrisnz:
Are the CI tests running? on my fork they only take 10~ minutes

Issue #527, opened 2021-03-08
node_modules included for SimpleSAMLphp?


sammarshallou:
Since the last update of the SimpleSAMLphp library in .extlib, the node_modules has been included. It wasn't included in the previous versions.

Is this actually needed or is it like Moodle where it's an artefact of the build process and doesn't need to be in the repo? Because it's 11k files, it has a non-negligible effect on deployments etc.

I tried deleting it on our system and I can still log in but I realise this is not a very thorough test and maybe there is some reason it's needed. :) However, in case not, I'll submit a pull request.

kabalin:
I think this should be closed as being addressed in #528 ?

danmarsden:
Yep

PR #528, opened 2021-03-08
Remove node_modules from SimpleSAMLphp #527

sammarshallou:


sammarshallou:
sorry I did the pull request wrong, I filed the issue first then the pull request, shouldn't have filed a separate issue! Anyway the description from the issue #527 is:

Since the last update of the SimpleSAMLphp library in .extlib, the node_modules has been included. It wasn't included in the previous versions.

Is this actually needed or is it like Moodle where it's an artefact of the build process and doesn't need to be in the repo? Because it's 11k files, it has a non-negligible effect on deployments etc.

I tried deleting it on our system and I can still log in but I realise this is not a very thorough test and maybe there is some reason it's needed. :) However, in case not, I'll submit a pull request.

brendanheywood:
thanks @samchaffee can you please also touch this with the removing step:

https://github.com/catalyst/moodle-auth_saml2/blob/master/.extlib/UPGRADE.md

danmarsden:
yep - I see the actual SimpleSAMLphp 1.18.8 release doesn't include the node_modules dir either - probably just accidentally included when doing the last simplesaml update. 

I see the 1.19 release doesn't have it either so I suspect we won't need the line in the upgrade.md but leaving it there to make sure we don't add it again.

PR #530, opened 2021-03-19
Match simplesaml session.duration to Moodle sessiontimeout

andrewhancox:
Backport fix from master branch


brendanheywood:
thanks @andrewhancox

PR #532, opened 2021-03-31
Adds role mapping from Idp

jvinolas:
We've implemented the admin, course creator and manager role mapping. Tested working against keycloak role mapping.


locallib.php
```diff
@@ -545,3 +545,32 @@ function auth_saml2_admin_nav($title, $url) {
...
+            $siteadmins[] = $user->id;
+            $newAdmins = implode(',', $siteadmins);
+            set_config('siteadmins', $newAdmins);
+        }
+    }
+
+    // Process coursecreator and manager
+    $syscontext = context_system::instance();
+    if(in_array($config->saml_role_coursecreator_map,$attributes['Role'])){
+        $creatorrole = $DB->get_record('role', array('shortname'=>'coursecreator'), '*', MUST_EXIST);
```
brendanheywood, line 569: I'd like for the roles to not be hard coded here, this should be configurable

lang/en/auth_saml2.php
```diff
@@ -210,3 +210,13 @@
 $string['regeneratepath'] = 'Certificate path path: {$a}';
 $string['regenerateheader'] = 'Regenerate Private Key and Certificate';
 $string['regeneratesuccess'] = 'Private Key and Certificate successfully regenerated';
+
+/*
+ * Role mapping
+ */
+$string['saml_role_map'] = "Role";
+$string['saml_rolemapping'] = "Role Mapping";
+$string['saml_rolemapping_head'] = "The IdP can use it's own roles. Set in this section the mapping between IdP and Moodle roles. Accepts multiple valued comma separated. Example: admin,owner,superuser.";
```
brendanheywood, line 219: I think I'd like to see a more clear mapping and example here, I think something like a set of key value pairs one per line, like

```
keycloakadmin,admin
keycloakmanager,coursemanager
keycloakstudent,learner
```

This would make it much more powerful and also easier to see what is actually happening. It should also validate against the configured moodle roles.

There is also the important question of negative constraints, ie if you had the admin role in keycloak and then it was removed, then we should be able to declare that this role is fully managed by saml and it should be removed if needed and not just added.

I'm not sure how to best express this in config but a simple idea is an optional 3rd column which says if this role should be removed if not present:

```
keycloakadmin,admin,remove
keycloakmanager,coursemanager
keycloakstudent,learner
```

One last thing, the key name of the 'Role' saml attribute should also configurable too., and it can default to 'Role'

brendanheywood:
hi @jvinolas this looks really great and this feature is well overdue. A few comments in the pr, and can you please also touch the README.md where it mentions roles not being managed?

lang/en/auth_saml2.php
brendanheywood: I can't see how this could work via cron, it can only be based on the latest attributes it gets during a particular login. So the idp says these are the roles you have right now, Moodle then takes those and if you don't have any attributes which map to a role which you do have then it removes it on the spot.

Also note we've now branches into MOODLE_39_STABLE can you please point this pr against it?

jvinolas: We are using this PR in production and we use keycloak as the only user source so our desired behaviour is like you say: it updates de role when the users logs in again if it was modified in keycloak.
Maybe there should be a checkbox to set or unset role updating at each user login.
We will try to point it to your latest branch and update the README next week.

jvinolas: It's updated to MOODLE_39_STABLE

kabalin:
Hi @jvinolas, just edit PR and make it target MOODLE_39_STABLE rather than master. Thanks!

danmarsden:
thanks @kabalin - I've just adjusted that - ideally I'd prefer @brendanheywood to come back and finish off the PR, but if not @nhoobin do you want to look at this one?

danmarsden:
hmm - I'd add a -1 to merging this with hard-coded role shortnames (coursecreator/manager) - I think we'd need to tidy up the way roles are mapped before merging this.

classes/auth.php
```diff
@@ -687,6 +687,9 @@ public function saml_login_complete($attributes) {
             set_config('siteadmins', implode(',', $admins));
         }
 
+        // Synchronize IdP roles to moodle
+        sync_roles($user, $attributes, $this->config);
```
kabalin, line 691: If `$attributes['Role']` is not defined, this function should not be called at all. Also, it worth moving it into `auth` class rather than keeping it in `locallib.php`, it is more logical given this function operates in context of authentication process and also you would not need to pass `$this->config` to it.

kabalin (changes requested):

kabalin:
With regard to role assigning, I would suggest using approach similar to one in `auth_ldap`: https://github.com/moodle/moodle/blob/master/auth/ldap/auth.php#L1838-L1859, so that the actual component that does role allocation is recorded and those roles do not interfere with other roles user may have. This will simplify role unassigning as well (those roles not listed in attribute will be removed as long as they were allocated through this plugin).

classes/auth.php
jvinolas: > With regard to role assigning, I would suggest using approach similar to one in `auth_ldap`: https://github.com/moodle/moodle/blob/master/auth/ldap/auth.php#L1838-L1859, so that the actual component that does role allocation is recorded and those roles do not interfere with other roles user may have. This will simplify role unassigning as well (those roles not listed in attribute will be removed as long as they were allocated through this plugin) and you would not need to hardcode roles (any assignable role will work). IMO site admin allocation though this is not a good idea, I would suggest removing this option.

We integrate moodle, nextcloud and wordpress with SAML and we use the admin role to just allow also admin to go from one app to the other and configure it. Is it a problem to just let admin role mapping there as an option if one wants to use it in his deployment? If not mapped it won't be used at all.

danmarsden: If you want to keep working on this and get it into a state that we could accept it might be cleaner to remove the admin role sync for now, tidy up the role mapping to remove the hard-coded shortnames and then once that is good enough to land you could look at a separate PR to add site-level admin mapping support.

Issue #416, opened 2020-04-17
Behat hook gives error on any Behat failure on Windows systems


sammarshallou:
On a Windows system, when you run Behat and any Behat scenario fails (not related to auth_saml2), if you have auth_saml2 installed, there is a spurious 'Failed hooks' error caused by auth_saml2 - even if the plugin isn't being tested and isn't even enabled.

It is caused by this function which attempts to take a screenshot and save it into /tmp:

https://github.com/catalyst/moodle-auth_saml2/blob/master/tests/behat/behat_auth_saml2.php#L47

I am not sure it should really be doing that as part of this plugin, since it seems to run for all Behat tests systemwide. Might be better to remove that function?

The actual error is:

--- Failed hooks:

    AfterStep # behat_auth_saml2::take_screenshot_after_failure()
      Warning: file_put_contents(/tmp/behat_screenshots.base64): failed to open stream: No such file or directory in C:\Users\(my user account)\workspace\ou-moodle2\auth\saml2\tests\behat\behat_auth_saml2.php line 56

timhunt:
$CFG->behat_faildump_path FTW. An individual plugin definitely should not be hacking this.

sammarshallou:
Thanks Dan for merging this, the issue should now be fixed.

PR #535, opened 2021-04-01
Removed Behat hook to take screenshot on failure #416

sammarshallou:
Removes the Behat hook that takes a screenshot on failure. When you have the plugin installed, this applies after every failure from every Behat file (even those not related to auth_saml2) and also it crashes on Windows because of hard-coded /tmp path. Since Behat already has a facility to take screenshots on failure anyway (which works on all platforms), the hook should simply be removed IMO.

### DIFF
--- a/tests/behat/behat_auth_saml2.php
+++ b/tests/behat/behat_auth_saml2.php
@@ -27,7 +27,6 @@
 use auth_saml2\admin\saml2_settings;
 use auth_saml2\task\metadata_refresh;
 use Behat\Behat\Hook\Scope\AfterStepScope;
-use Behat\Mink\Exception\UnsupportedDriverActionException;
 use Behat\Mink\Exception\ExpectationException;
 use Behat\Gherkin\Node\TableNode;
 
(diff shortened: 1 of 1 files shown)


danmarsden:
seems to make sense - I suspect we need to remove this line at the same time:
https://github.com/catalyst/moodle-auth_saml2/blob/master/tests/behat/behat_auth_saml2.php#L30

any chance you could update the patch to remove that as well?

thanks!

danmarsden:
seems to make sense - I suspect we need to remove this line at the same time:
https://github.com/catalyst/moodle-auth_saml2/blob/master/tests/behat/behat_auth_saml2.php#L30

any chance you could update the patch to remove that as well?

thanks!

sammarshallou:
Thanks! You're right, that line doesn't appear to be used now, sorry I missed it before. I've removed it and pushed the new commit. Weird thing, it passes in my repository https://github.com/sammarshallou/moodle-auth_saml2/actions so I think it is fine, but it still shows as 'haven't completed yet' in this PR...

danmarsden:
Thanks Sam - github actions config was wrong - I've just fixed that now and it looks like it should trigger the tests here - will see if they pass and will then merge this in - thanks!

Issue #537, opened 2021-04-14
Core files phpunit is failing because of autologin support


golenkovm:
Hi team,

I've just found a failing unit test:

```
root@f5f1a3f62cb1:/siteroot# vendor/bin/phpunit files/tests/externallib_test.php --stop-on-failure
Moodle 4.0dev (Build: 20210330), f0eb6a5729b64318bd9d794e440ac400b9240602
Php: 7.4.3, mysqli: 5.7.31, OS: Linux 5.4.0-70-generic x86_64
PHPUnit 9.5.2 by Sebastian Bergmann and contributors.

E

Time: 00:00.139, Memory: 60.50 MB

There was 1 error:

1) externallib_test::test_upload
Undefined index: REQUEST_METHOD

/siteroot/auth/saml2/classes/auto_login.php:67
/siteroot/auth/saml2/lib.php:58
/siteroot/lib/moodlelib.php:2829
/siteroot/lib/externallib.php:515
/siteroot/files/externallib.php:339
/siteroot/files/tests/externallib_test.php:62
/siteroot/lib/phpunit/classes/advanced_testcase.php:80

ERRORS!
Tests: 1, Assertions: 1, Errors: 1.
```

This seems to be a regression of https://github.com/catalyst/moodle-auth_saml2/commit/6b9de9a3e7ac0b0f174f201eebbd57ba61373abe

Peterburnett:
Also seeing this on another test:

```
1) core\external\record_userfeedback_action_testcase::test_record_userfeedback_action with data set "give action" ('give')
Undefined index: REQUEST_METHOD
```

danmarsden:
this was resovled in #538 - thanks!

PR #538, opened 2021-04-14
Fix core files unit test

golenkovm:
This is a fix for #537

### DIFF
--- a/classes/auto_login.php
+++ b/classes/auto_login.php
@@ -64,7 +64,7 @@ class auto_login {
         }
 
         // If this is not a GET request, don't try autologin.
-        if ($_SERVER['REQUEST_METHOD'] !== 'GET') {
+        if (empty($_SERVER['REQUEST_METHOD']) || $_SERVER['REQUEST_METHOD'] !== 'GET') {
             return;
         }
 


golenkovm:
With the patch in place the testcase is passing:
```
root@f5f1a3f62cb1:/siteroot# vendor/bin/phpunit files/tests/externallib_test.php                  
Moodle 4.0dev (Build: 20210330), f0eb6a5729b64318bd9d794e440ac400b9240602
Php: 7.4.3, mysqli: 5.7.31, OS: Linux 5.4.0-70-generic x86_64
PHPUnit 9.5.2 by Sebastian Bergmann and contributors.

......                                                              6 / 6 (100%)

Time: 00:01.416, Memory: 66.50 MB

OK (6 tests, 15 assertions)
```

golenkovm:
Hey @danmarsden 

Could you please have a look at this PR once you have a chance.

Kind regards,
Mikhail

danmarsden:
Thanks @golenkovm 

@sammarshallou - do you have any thoughts here? - I'm not convinced the patch is right - if request_method is not set at all, shouldn't we be ignoring the login check as well?

eg maybe something like:
```
if (empty($_SERVER['REQUEST_METHOD']) || $_SERVER['REQUEST_METHOD'] !== 'GET')

sammarshallou:
hi @danmarsden - yes I agree with you, your version of the fix looks correct to me, thanks (sorry for causing the bug)

danmarsden:
Thanks @sammarshallou - @golenkovm could you please update the PR with that change?

golenkovm:
Thanks @danmarsden
The patch is amended.
Sorry for not looking at it in details.
Cheers

PR #541, opened 2021-04-29
Improve user search in saml_login_complete method.

juancs:
You should add mnethostid as a condition in order to benefint from the "username" index in "user" table... unless you have a reason for not doing that ;)


juancs:
Closed accidentaly :)

Issue #529, opened 2021-03-10
remove loginhttps handling from main branch


danmarsden:
We should remove these from the code in the main branch - loginhttps is no longer supported:

https://github.com/catalyst/moodle-auth_saml2/blob/f2ec4cea2115449af7e59a6c628887b4e1e64994/config/config.php#L33
https://github.com/catalyst/moodle-auth_saml2/blob/379be506d3552ddd14fa2c01c4d71638e4c0c9be/config/authsources.php#L33

brendanheywood:
+1

andyjdavis:
Hi. I have put up a PR that I believe takes care of this https://github.com/catalyst/moodle-auth_saml2/pull/544

Let me know if I haven't understand what you were wanting to achieve :)

danmarsden:
thanks! - merged!

PR #544, opened 2021-05-03
Removing references to the now unsupported loginhttps

andyjdavis:
This is a potential fix for https://github.com/catalyst/moodle-auth_saml2/issues/529


danmarsden:
yep - that's it - thanks @andyjdavis - merged!

PR #548, opened 2021-05-20
Code improvements

kabalin:
This patch contains a list of changes that makes plugin in a better position for further improvements. Suggested changes do not affect its functionality, i.e. some existing warnings remain in place, more changes will come after in separate PRs.

1. Moving auth class to classes directory - this makes easier to access class using namespace, and also improves structure. It also introducing `api` class for storing static methods not directly related to `auth` class (also not to overload the latter with functions).
2. Move static `saml2_authproc_filters_hook` to `api` class.
3. Refactor standalone test pages (`test.php` and `testing.php`)
    - Make them available only if debugging mode (`auth_saml2_debug=true`) is on, it is not good to have them available on production sites.
    - Requires auth plugin to be enabled.
    - Fix warning and codechecker issue.
4. Improve debugging - replaces `error_log` with `debugging` calls where appropriate, also it fixes and expand related unittest.
5. Unitest improvements - addresses codechecker issues, fixes some phpunit warnings.
6. Minor changes to GHA workflows - node is installed automatically, no need to add it to matrix, also Moodle 4 changes.


kabalin:
As suggestion, I it is probably safe to remove one of DBs from matrix, that would make CI test complete faster. There are no complex DB queries used in plugins, in fact DB interaction is minimal, it is very unlikely that it may pass with one of DBs but fail with the other.

tester.php
```diff
@@ -22,9 +22,23 @@
...
-// @codingStandardsIgnoreStart
+// Require_login is not needed here.
+// phpcs:disable moodle.Files.RequireLogin.Missing
 require_once(__DIR__ . '/../../config.php');
-// @codingStandardsIgnoreEnd
+require_once(__DIR__ . '/locallib.php');
+
+// Check we are in debug mode to use this tool.
+$saml2auth = new \auth_saml2\auth();
+if (!$saml2auth->is_debugging()) {
```
brendanheywood, line 32: I'm not sure I want to disable this test page if debugging is off. For example I only ever turn debugging on if there is a problem I need to debug, but I use the testing page all the time when configuring it. Is there some security concern you have?

auth.php
```diff
@@ -21,1148 +21,17 @@
...
-
-    /**
-     * Returns a list of potential IdPs that this authentication plugin supports.
-     * This is used to provide links on the login page.
-     *
-     * @param string $wantsurl the relative url fragment the user wants to get to.
-     *
-     * @return array of IdP's
-     */
-    public function loginpage_idp_list($wantsurl) {
```
brendanheywood, line 203: This function cannot be moved its a callback that core calls

brendanheywood: Actually no I see its moved to parent class, but is there a good reason this and the other sibling methods that are called by core should be in the parent rather than in this one? ie should is_internal, is_configured, loginpage_hook etc stay here?

classes/api.php
```diff
@@ -0,0 +1,112 @@
...
+
+defined('MOODLE_INTERNAL') || die();
+
+use moodle_url;
+
+/**
+ * Static list of api methods for auth saml2 configuration.
+ *
+ * @package   auth_saml2
+ * @copyright 2021 Moodle Pty Ltd <support@moodle.com>
```
brendanheywood, line 27: If things are mostly just a refactor of existing code can you please add or retail copyright for 'Catalyst IT' please

auth.php
kabalin: Just to be clear, the core convention auth.php located at root now looks like:
```
class auth_plugin_saml2 extends \auth_saml2\auth {
}
```
So, all methods that core is expecting to be there are reachable. We have got core `auth_oauth2` plugin that uses the same appraoch, it is just more convenient. If at any other place the plugin is referenced by child, it will not fail either (e.g. in unittest it its instance is requested as `get_auth_plugin('saml2')` and all works).

tester.php
kabalin: Yep, I have slight security concerns that user may get info there she is not supposed to see. Notice it is not checking moodle DEBUG_DEVELOPER mode, it is checking if Debugging is enabled in the plugin settings only:
image

It is generally not good exposing test pages on production sites that are available for general public.

classes/api.php
kabalin: OK

kabalin:
Added another commit that relocates PHPUnit test files according to convention (https://docs.moodle.org/dev/PHPUnit_integration#Organisation_of_test_files)

classes/api.php
```diff
@@ -0,0 +1,56 @@
...
+
+defined('MOODLE_INTERNAL') || die();
+
+use moodle_url;
+
+/**
+ * Static list of api methods for auth saml2 configuration.
+ *
+ * @package   auth_saml2
+ * @copyright Brendan Heywood <brendan@catalyst-au.net>
```
brendanheywood, line 27: This was probably wrong in the source but ongoing we need our copyright to be like this:

```
 * @author Brendan Heywood <brendan@catalyst-au.net>
 * @copyright Catalyst IT
 ```
so where you want dual copyright I think you want something like this:
```
 * @author Brendan Heywood <brendan@catalyst-au.net>
 * @copyright Catalyst IT
 * @author Ruslan ....
 * @copyright Moodle.com
```

kabalin: Changed accordingly 👍

Issue #499, opened 2021-01-08
Logging back in after logging out with Attempt IdP Signout set to No seems to allow you to log back in without checking the IdP for your session state


jstilwell:
First of all, I love your plugin. Thank you for maintaining it.

I discovered a fairly weird problem that I think may be a bug.

If you login to Moodle via SAML and then log out with Attempt IdP Signout set to No, when you click the SAML login button, it signs you back in. This is expected behavior.  However, if you manually log out of your IdP and then click Moodle's SAML login button again, it will still log you right back in.

If you do this with a SAML tracer extension on, you will see that when you click the SAML login button, there's no communication back to the IdP.

What I discovered is  that somehow your authenticated state is bound to the browser state in some way I haven't found yet, because if you logout of Moodle again, close the browser completely and reopen it, and then click to login to Moodle via SAML, it will redirect you to the IdP.

This seems like a substantial bug for scenarios where a computer may be shared by multiple users.

brendanheywood:
There are 3 sessions, the idp session, the sp session, and the moodle session. The sp one in the middle is mostly invisible but I think is the issue here. When that setting is off it logs you out of moodle, but does not attempt to log you out of the sp or the idp. But it probably should still log you out of the sp. As a quick test, reproduce it and then just before you try the re-login, look at the cookies and delete the MDL_SSP_SessID cookie which will kill the middle session.

Probably the easiest approach is manually kill that cookie on moodle logout when the setting is off. Another approach is if that setting is off then the session length for the SP session can be very short. Pull requests welcome or if you'd like to sponsor this fix please contact us and reference this issue https://www.catalyst-au.net/contact-us/sydney#email-us

badgerhill:
Had the same issue. Easy fix
in auth.php look for  public function logoutpage_hook() 

after line 1003 
1002 // Only log out of the IdP if we logged in via the IdP. TODO check session timeouts.
1003       if ($auth->isAuthenticated()) {
add these lines to delete the cookie

` $myCookie = 'MOODLEIDP1_'.$CFG->sessioncookie;`
`setcookie($myCookie, '', time() - HOURSECS, $CFG->sessioncookiepath, $CFG->sessioncookiedomain,
  $cookiesecure, $CFG->cookiehttponly);`

brendanheywood:
thanks @badgerhill - are you able to sort that into a  pull request please?

jstilwell:
[...]

PR #549, opened 2021-05-24
499 Delete session cookie on logout

jstilwell:
User badgerhill provided a bug fix for issue 499 but never provided a PR, so I'm doing it on their behalf.

Deleting the session cookie on IdP logout after checking to see if the user is authenticated via the IdP will fix the problem outlined in #499, wherein a user can logout, but still access the Moodle site.

### DIFF
--- a/auth.php
+++ b/auth.php
@@ -1003,6 +1003,11 @@ class auth_plugin_saml2 extends auth_plugin_base {
 
         // Only log out of the IdP if we logged in via the IdP. TODO check session timeouts.
         if ($auth->isAuthenticated()) {
+            
+            // Delete session cookie to address security concern in issue 499
+            $myCookie = 'MOODLEIDP1_'.$CFG->sessioncookie;
+            setcookie($myCookie, '', time() - HOURSECS, $CFG->sessioncookiepath, $CFG->sessioncookiedomain, $cookiesecure, $CFG->cookiehttponly);
+
             $this->log(__FUNCTION__ . ' Do SSP logout');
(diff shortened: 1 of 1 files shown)


auth.php
```diff
@@ -1003,6 +1003,11 @@ public function logoutpage_hook() {
 
         // Only log out of the IdP if we logged in via the IdP. TODO check session timeouts.
         if ($auth->isAuthenticated()) {
+            
+            // Delete session cookie to address security concern in issue 499
+            $myCookie = 'MOODLEIDP1_'.$CFG->sessioncookie;
+            setcookie($myCookie, '', time() - HOURSECS, $CFG->sessioncookiepath, $CFG->sessioncookiedomain, $cookiesecure, $CFG->cookiehttponly);
```
kabalin, line 1009: I believe $CFG and $cookiesecure are not defined here. Did you test the fix?

You can delete cookie by calling:
```
$this->set_idp_cookie('');
```

kabalin (commented):
I am not sure about this change in fact, may be maintainers can comment. There is no security concern, this cookie is used to select IdP in dropdown menu on `auth/saml2/selectidp.php` if "Remember login service" was checked, it stores last IdP selection. I think we don't need to do anything about it if we need "remember login servise" functionality to work.

Cross referencing #499

danmarsden:
Thanks @kabalin - I agree - this patch is modifying the wrong thing - closing.

brendanheywood:
This was a real issue and I've got a fix here:

https://github.com/catalyst/moodle-auth_saml2/pull/651

PR #553, opened 2021-05-28
Limit mariadb GH workflow to master only.

kabalin:


dmitriim:
I'd give -1 to this change. I had some code issues that were caught by CI running on mariadb.

kabalin:
Per discussion with @danmarsden, limiting mariadb run to master only.

kabalin:
I think half of those tests having node env defined (14.15.1), but how they propagate here, that is not in master any more (workflows with node version).

danmarsden:
Thank @dmitriim - I had the same thought. I did think we could probably drop down to a single check on mariadb which Ruslan has pushed now - but it would be good for someone else to +1 this before we merge. We want to be good citizens using actions and only test useful things.

danmarsden:
No one else has objected... so I'm going to merge this in - happy for someone else to object and we can add some mariadb checks back in though.

PR #556, opened 2021-05-28
Improve auth::loginpage_idp_list

### LINKED ISSUE #555: loginpage_idp_list metadata configuration check does not take into account entity active flag.
If there are more than one IdPs exist, but only one is active, the Moodle login page outputs:

>Missing identity configuration for 'http://samlidp.local:8080/simplesaml/saml2/idp/metadata.php': Please check/save SAML2 configuration or if able to inspect the database, check: SELECT * FROM {auth_saml2_idps} WHERE metadataurl='http://samlidp.local:8080/simplesaml/saml2/idp/metadata.php' -- Remember to purge caches if you make changes in the database.

Patch is ready and will follow.

kabalin:
* Add unit tests.
* Fix "missing entity record" warning when one of the entities is inactive.

Closes #555

### DIFF
--- a/classes/auth.php
+++ b/classes/auth.php
@@ -202,71 +202,65 @@ class auth extends \auth_plugin_base {
         // The array of IdPs to return.
         $idplist = [];
 
-        foreach ($this->metadatalist as $metadata) {
-            if (!array_key_exists($metadata->idpurl, $this->metadataentities)) {
-                $message = "Missing identity configuration for '{$metadata->idpurl}': " .
-                           'Please check/save SAML2 configuration or if able to inspect the database, check: ' .
-                           "SELECT * FROM {auth_saml2_idps} WHERE metadataurl='{$metadata->idpurl}' " .
-                           '-- Remember to purge caches if you make changes in the database.';
-                debugging($message);
+        // Create IdP metadata url => name mapping.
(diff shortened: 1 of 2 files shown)


danmarsden:
I see this is now failing on the 3.5 tests - although we could fix this... I'm inclined to branch saml2 - @brendanheywood thoughts?

kabalin:
Let me know what you decide. I would suggest to branch and remove 3.5 CI from master.

danmarsden:
Yeah - I'd be more inclined to branch for 3.9 and higher, creating a new branch that supports 3.9 and higher and leaving the older branch for any old (unsupported Moodle) releases or Moodle forks that are still needing this older version.

@brendanheywood are you happy to give a +1 to branch for Moodle 3.9 ? - this would involve us going through that new branch process:

* Creating a new MOODLE_39_STABLE branch for all newer releases and setting that as the default branch.
* Flagging master branch as deprecated (readme/default branch etc)- and considering the addition of a $plugin->supported = [35, 39]; to the version.php file on master.(need to check how Totara will deal with this - we may need to leave it out.)
* Creating a new MOODLE_35_STABLE branch which is a copy of the existing master branch - exluding this patch here as it is not backwards compatible.
* Create internal WR for AU team to update saml2 pipelines based on master branch.
* When AU team have resolved submodule pipelines, delete the old master branch.

brendanheywood:
+1 @danmarsden that pretty well exactly what I would have pitched

PR #568, opened 2021-06-15
Trigger user_login_failed event on unsucessful login.

kabalin:
This depends on #562.

### DIFF
--- a/classes/auth.php
+++ b/classes/auth.php
@@ -596,10 +596,8 @@ class auth extends \auth_plugin_base {
                 $this->log(__FUNCTION__ . " to lowercase for $uid");
                 $uid = strtolower($uid);
             }
-            if ($user = $DB->get_record('user', array(
-                    $this->config->mdlattr => $uid,
-                    'deleted' => 0,
-                    'mnethostid' => $CFG->mnet_localhost_id))) {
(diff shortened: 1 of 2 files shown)


classes/auth.php
```diff
@@ -573,42 +573,56 @@ public function saml_login_complete($attributes) {
...
-        if (empty($attributes[$attr]) ) {
+        if (empty($attributes[$attr])) {
+            // Missing mapping IdP attribute. Login failed.
+            $event = \core\event\user_login_failed::create(['other' => ['username' => 'unknown',
+                'reason' => AUTH_LOGIN_NOUSER]]);
+            $event->trigger();
             $this->error_page(get_string('noattribute', 'auth_saml2', $attr));
         }
 
+        // Testing user's groups and allow access according to preferences.
```
danmarsden, line 584: Shouldn't this access check happen "After" we have worked out if we have a valid user?

kabalin: It does not need user record, it checks group in the attributes against plugin configuration.

kabalin: Possibly the comment can be improved.

danmarsden: of course - so in fact it makes a lot more sense to happen before we check if the user exists in Moodle - much better where you've put it there.

PR #585, opened 2021-07-13
Allow configuration of more SAML settings

alexmorrisnz:
Adds the ability to configure:
- Entity ID
- 'WantAssertionsSigned'
- ACS Bindings
- 'AllowCreate'

Allows you to match users on the nameid value

This allows us to use RealMe with auth_saml2


classes/auth.php
```diff
@@ -594,14 +596,19 @@ public function saml_login_complete($attributes) {
...
-        foreach ($attributes[$attr] as $uid) {
-            if ($this->config->tolower) {
-                $this->log(__FUNCTION__ . " to lowercase for $uid");
-                $uid = strtolower($uid);
-            }
-            if ($user = get_complete_user_data($this->config->mdlattr, $uid)) {
-                // We found a user.
-                break;
+        if ($this->config->matchnameid) {
+            $uid = $nameid->value;
```
kabalin, line 600: Would be good to have test covering this case (the rest of `saml_login_complete` is test covered already).

lang/en/auth_saml2.php
```diff
@@ -108,6 +108,16 @@
...
+$string['wantassertionssigned'] = 'Want Assertions Signed';
+$string['wantassertionssigned_help'] = '';
+$string['assertionsconsumerservices'] = 'Assertions Consumer Services';
+$string['assertionsconsumerservices_help'] = '';
+$string['entityid'] = 'Entity ID';
+$string['entityid_help'] = '';
+$string['allowcreate'] = 'Allow Create';
+$string['allowcreate_help'] = '';
+$string['matchnameid'] = 'Match on NameID value';
+$string['matchnameid_help'] = '';
```
brendanheywood, line 120: I think this definitely needs help text to distinguish it from nameidasattribute and explain when you would use one over the other, and possibly move them to be adjacent

lang/en/auth_saml2.php
```diff
@@ -108,6 +108,16 @@
...
 $string['mdlattr_help'] = 'Which Moodle user field should the IdP attribute be matched to?';
 $string['mdlattr'] = 'Mapping Moodle';
+$string['wantassertionssigned'] = 'Want Assertions Signed';
+$string['wantassertionssigned_help'] = '';
+$string['assertionsconsumerservices'] = 'Assertions Consumer Services';
+$string['assertionsconsumerservices_help'] = '';
+$string['entityid'] = 'Entity ID';
+$string['entityid_help'] = '';
+$string['allowcreate'] = 'Allow Create';
+$string['allowcreate_help'] = '';
```
brendanheywood, line 118: This definitely needs docs to tell it apart from 'autocreate' and I think both the autocreate and these docs should be refined something like:

allowcreate Allow creation of IdP users on demand
autocreate Allow creation of Moodle users on demand

settings.php
```diff
@@ -165,6 +165,54 @@
...
+        'urn:oasis:names:tc:SAML:1.0:profiles:browser-post' => 'Browser post profile',
+        'urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Artifact' => 'HTTP Artifact',
+        'urn:oasis:names:tc:SAML:1.0:profiles:artifact-01' => 'Artifact 01 profile',
+    ];
+
+    $settings->add(new admin_setting_configmultiselect(
+        'auth_saml2/assertionsconsumerservices',
+        get_string('assertionsconsumerservices', 'auth_saml2'),
+        get_string('assertionsconsumerservices_help', 'auth_saml2'),
+        [
```
brendanheywood, line 193: worth a micro refactor of array_keys($assertionsconsumerservices) ?

lang/en/auth_saml2.php
```diff
@@ -108,6 +108,16 @@
 $string['manageidpsheading'] = 'Manage available IdPs';
 $string['mdlattr_help'] = 'Which Moodle user field should the IdP attribute be matched to?';
 $string['mdlattr'] = 'Mapping Moodle';
+$string['wantassertionssigned'] = 'Want Assertions Signed';
+$string['wantassertionssigned_help'] = '';
+$string['assertionsconsumerservices'] = 'Assertions Consumer Services';
+$string['assertionsconsumerservices_help'] = '';
+$string['entityid'] = 'Entity ID';
+$string['entityid_help'] = '';
```
brendanheywood, line 116: gonna sound like a broken record - this needs docs, none of these should be empty. This one I think its quite important to tell people that in 99.9% of cases they can leave it blank

lang/en/auth_saml2.php
alexmorrisnz: I wasn't aware of nameidasattribute, seems to work fine with realme so will remove this setting

classes/auth.php
alexmorrisnz: Removed this option as nameidasattribute covers it

settings.php
alexmorrisnz: yep, done

lang/en/auth_saml2.php
alexmorrisnz: yep added the help texts, should this be documented anywhere else?

lang/en/auth_saml2.php
```diff
@@ -108,6 +108,14 @@
 $string['manageidpsheading'] = 'Manage available IdPs';
 $string['mdlattr_help'] = 'Which Moodle user field should the IdP attribute be matched to?';
 $string['mdlattr'] = 'Mapping Moodle';
+$string['wantassertionssigned'] = 'Want Assertions Signed';
```
danmarsden, line 111: Moodle uses sentence case - not camel case - also check the other strings like consumer services...

settings.php
```diff
@@ -165,6 +165,42 @@
             get_string('spmetadatasign_help', 'auth_saml2'),
             0, $yesno));
 
+    $settings->add(new admin_setting_configtext(
+        'auth_saml2/entityid',
+        get_string('entityid', 'auth_saml2'),
+        get_string('entityid_help', 'auth_saml2'),
+        "$CFG->wwwroot/auth/saml2/sp/metadata.php"
```
danmarsden, line 172: shouldn't the default be blank? - eg what happens when a prod site gets set up and then sync to test... you end up with a testing site pointing at prod site metadata.php accidentally....

lang/en/auth_saml2.php
alexmorrisnz: done

settings.php
alexmorrisnz: yep, set to '' now

alexmorrisnz:
I've added a new setting 'AuthnContext'

alexmorrisnz:
Fixed the behat tests

lang/en/auth_saml2.php
```diff
@@ -108,6 +108,16 @@
 $string['manageidpsheading'] = 'Manage available IdPs';
 $string['mdlattr_help'] = 'Which Moodle user field should the IdP attribute be matched to?';
 $string['mdlattr'] = 'Mapping Moodle';
+$string['wantassertionssigned'] = 'Want assertions signed';
+$string['wantassertionssigned_help'] = 'Whether assertions received by this SP must be signed';
+$string['assertionsconsumerservices'] = 'Assertions consumer services';
+$string['assertionsconsumerservices_help'] = 'List of bindings the SP should support';
+$string['entityid'] = 'Entity ID';
+$string['entityid_help'] = 'Unique name of the entity provider. Leave blank unless required';
+$string['allowcreate'] = 'Allow Create';
```
danmarsden, line 117: camel case again :-)

danmarsden:
@alexmorrisnz - I'm not sure about the way you're setting the acs.Bindings default values?

I note that the simplesamlphpdocs list 5 options:
https://github.com/simplesamlphp/simplesamlphp/blob/5a705f4dba77b8aa37d3600313b49f6c970b1ef4/modules/saml/docs/sp.md#options

but you've only added 4 and you've set them all to be used by default which would potentially stop the 5th option from workingno idea if that mising one is used now, but if not, I'm guessing it might be used in future?

I'm tempted to suggest we leave the default value for this blank (nothing selected) and then only set the acs.Bindings param when that value is not empty - that way the library will be able to use "all" available bindings rather than just the ones set in the list - this might be important in future if simplesaml upstream library adds new binding options too.

alexmorrisnz:
Hi @danmarsden thanks, I wasn't aware of the 5th. I've added it as an option now
I agree with leaving it blank, makes sense. I've made the change

alexmorrisnz:
Rebased on the latest MOODLE_39_STABLE branch

PR #587, opened 2021-07-20
Add type for return value same as interface

TomoTsuyuki:


danmarsden:
grr bad @kabalin ;-)

kabalin:
Thanks for fixing this guys and sorry for introducing regression.

PR #589, opened 2021-07-21
Pass an empty string instead of a null value when there is no IDP sesion value

### LINKED ISSUE #588: Unit Test Failure in MWP 3.11.1 - auth_test.php > null instead of string given
When running the unit test suite for this plugin when installed in a MWP 3.11.1 site test throw errors.

Error received is:
`TypeError: Argument 1 passed to tool_tenant\local\auth\saml2\manager::complete_login_hook() must be of the type string, null given`

"Full" stack trace: 
```
/var/www/workplace/xxxxxxxxxxx/manager.php:78
/var/www/workplace/lib/moodlelib.php:8000
/var/www/workplace/auth/saml2/classes/auth.php:616
/var/www/workplace/auth/saml2/tests/auth_test.php:476
/var/www/workplace/lib/phpunit/classes/advanced_testcase.php:80
```

There is a callback in `classes/auth.php` that passes a null in certain situations, code below:
```        component_class_callback('\tool_tenant\local\auth\saml2\manager', 'complete_login_hook',
            [$SESSION->saml2idp ?? null, $uid, $user]);
```

However the callback signature expects that value to be a string.  An easy fix seem to be passing an empty string instead.
I'll make a patch that does this

mattporritt:
As the callback signature expects a string type value

Fixes #588

### DIFF
--- a/classes/auth.php
+++ b/classes/auth.php
@@ -613,7 +613,7 @@ class auth extends \auth_plugin_base {
 
         // Moodle Workplace - Check IdP's tenant availability, for new user pre-allocate to tenant.
         component_class_callback('\tool_tenant\local\auth\saml2\manager', 'complete_login_hook',
-            [$SESSION->saml2idp ?? null, $uid, $user]);
+            [$SESSION->saml2idp ?? '', $uid, $user]);
 
         $newuser = false;
         if (!$user) {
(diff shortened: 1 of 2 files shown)


dmitriim:
I can confirm that the patch fixes unit tests in 3.11.1

dmitriim (approved):

Issue #582, opened 2021-07-02
GCMEncryption


lselanne:
GCMEncryption would need to be added as part of SAML2 authentication plugin.

The GCMEncryption is currently not available in the SAML2 plugin, but according to https://github.com/simplesamlphp/saml2/issues/179 discussion, it has been added as part of simplesamlphp.

More information: https://wiki.shibboleth.net/confluence/display/IDP4/GCMEncryption

danmarsden:
We are currently using v1.18.8 and it sounds like that landed in v 1.19 and higher - we should update to a newer library at some point, but there is a bit of work involved there in getting that right and testing it - feel free to reach out privately if you would like to fund the work required.

danmarsden:
@lselanne - if you were able to test the pull request #604 which contains an update to simplesaml to see if it fixes your issue that would be great!

lselanne:
> @lselanne - if you were able to test the pull request #604 which contains an update to simplesaml to see if it fixes your issue that would be great!

Great news. Unfortunately I can't test this since it would need many third parties involvement.

danmarsden:
I'm closing this off as it relates to simplesamlphp which we have since upgraded - feel free to report another issue if you have issues with saml2's implementation of simplesamlphp

PR #602, opened 2021-09-07
Cookie samesite config 39

melnox:
Allows a Moodle config option for samesite attribute with useragent checks for the support of 'none'

### DIFF
--- a/.extlib/simplesamlphp/lib/SimpleSAML/Utils/HTTP.php
+++ b/.extlib/simplesamlphp/lib/SimpleSAML/Utils/HTTP.php
@@ -17,6 +17,49 @@ use SimpleSAML\XHTML\Template;
 class HTTP
 {
     /**
+     * Determine if the user agent can support cookies being sent with SameSite equal to "None".
+     * Browsers with out support may drop the cookie and or treat is a stricter setting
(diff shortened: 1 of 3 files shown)


danmarsden:
hmm... not a fan of this landing in the simplesamlphp lib - have you sent this upstream to simplesamlphp or is this a copy of something done upstream?

danmarsden:
it's hard enough pulling in updates from simplesamlphp as it is now, without adding more changes to that lib.

danmarsden:
Looks like it's been copied from upstream:
https://github.com/simplesamlphp/simplesamlphp/blob/master/lib/SimpleSAML/Utils/HTTP.php#L19

I think we're better off doing a full simplesamlphp lib update to help fix other stuff like #582 rather than backporting individual changes from newer simplesamlphp lib - there are some docs here on how to do this...
https://github.com/catalyst/moodle-auth_saml2/blob/MOODLE_39_STABLE/.extlib/UPGRADE.md

but I might be able to find some time to do that if you don't have time yourself.

melnox:
Ya this was picked from upstream and then our config changed to accommodate the setting. I will have a look at doing the full upgrade for this rather then.

melnox:
I will close this off in favor of the upgrade which contains the patch anyway and create a new PR for the config file change.

PR #604, opened 2021-09-14
Simplesamlphp 1.19.1 update

melnox:


danmarsden:
thanks @melnox  - @brendanheywood / @kabalin - I think that includes all the custom changes we had on top of saml2 but would be good if you guys could sanity check that too?

@melnox - I suspect you have copied the 19 lib on top of the existing code rather than deleting the folder and recreating it with only the new library code? - what's that .idea folder doing in there in our main folder! - we should be removing that!! :-)

melnox:
@danmarsden yes I seem to have copied the wrong directory across. I will update this shortly.

danmarsden:
thanks @melnox that looks a bit more like what I would have expected to see - @kabalin any chance you might be able to test this on your end before we land it too?

thanks,

kabalin:
Hi @Nixon-cat, thanks for your work on this. I am confused why it includes some existing commits under different hashes?

danmarsden:
We usually replace the code with the new simplesaml release and cherry pick the custom changes again on top so that it's easy to see what our custom changes are.

kabalin:
> We usually replace the code with the new simplesaml release and cherry pick the custom changes again on top so that it's easy to see what our custom changes are.

Aha, did not realise they are ext library changes only, OK then.

melnox:
There were a few conflicts that needed to be resolved when cherry picking these across.

kabalin:
Hi, I confirm this branch does not break any known Workplace functionality. 👍 for merging.

PR #606, opened 2021-09-22
Add username to data mapping to close #605

### LINKED ISSUE #605: Add a way to map username to any of SAML attributes
Working on https://github.com/catalyst/moodle-auth_saml2/issues/577 I have realised that we set a username as value of "idpattr" of SAML data. Would be good to have this opportunity as in  my case username is different field in SAML feed.

dmitriim:


danmarsden:
makes sense to me - nice to see the check preventing the field being updated when it's used as the mapping field - merging.

PR #611, opened 2021-09-30
Issue#469: Remove PHPMailer from autoloaded files and composer files …

Anupamatd:
…from compiled extlib/simplesamlphp


danmarsden:
is it worth adding a unit test or some other form of test to prevent us hitting this again next time we do the update? - especially as we've hit it twice so far?

dmitriim:
yeah. Good point @danmarsden A quick test for the file presence as well as  scanning for PHPMailer in auto-lading files would help the next time I guess,

Anupamatd:
Sure @danmarsden @dmitriim - will add that. Thank you

Anupamatd:
Hi @dmitriim @danmarsden - added PHP unit test to cover that. Please review. Thank you

tests/phpmailer_test.php
```diff
@@ -0,0 +1,68 @@
...
+
+defined('MOODLE_INTERNAL') || die();
+
+/**
+ * auth_saml2 PHPMailer unit tests
+ *
+ * @package    auth_saml2
+ * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+class auth_saml2_phpmailer_testcase extends advanced_testcase {
```
dmitriim, line 32: In out particular case we are trying to check PHPMailer, however, I'm thinking if potentially we can add more checks that can be run after we simplesamlphp upgrade library without creating a test file for each of the case. 

Do you think it makes sense to rename this test file to something a bit more generic? E.g. auth_saml2_simplesamlphp_upgrade_testcase or something...

Anupamatd: Thank you @dmitriim for suggesting that !! Renamed file and class name. Please review. Thank you..

PR #613, opened 2021-10-09
 SOAP binding / backend logout work in progress proposal

schlupmann:
This is a work in progress / experimental proposal for a backend channel logout. It works in our environment.

I understand that any added code and the call to the specific logout function needs to be pulled out from the simplesamphp library. But to be able to get sessionids and register a logouthandler in sp/saml2-logout.php, we need to parse the xml message and pull quite a bit of the code from the extlib saml2-logout.php and logoutStore.php into sp/saml2-logout.php. It somehow defeats the purpose of keeping the plugin independent from the simplesamphp library.

I will keep on looking for a better solution.

(and sorry for the earlier mess... please remove the unnecessary zip files i posted on the subject)

### DIFF
--- a/.extlib/simplesamlphp/lib/SimpleSAML/Session.php
+++ b/.extlib/simplesamlphp/lib/SimpleSAML/Session.php
@@ -736,12 +736,9 @@ class Session implements \Serializable, Utils\ClearableState
         assert(isset($this->authData[$authority]));
 
         if (empty($this->authData[$authority]['LogoutHandlers'])) {
-            if (empty($this->authData[$authority]['Attributes']['username'][0])) {
-                return;
-            } else {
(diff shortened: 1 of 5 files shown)


danmarsden:
no worries - can be hard to work out how to do this the first time round! - nice work on that patch it's much easier to review in that form - thanks! - I'll leave @brendanheywood to comment further.

amayard:
Hello,

We submitted our take at the SOAP binding issue for the auth_saml2 plugin as a pull request to this fork.
https://github.com/schlupmann/moodle-auth_saml2/pull/1/commits/6bd71f85b760b5f5f91b1a58a890bad5ed747bef

We strongly based it on this proposal and payed attention to the reviewers comment. Doing so we tried to limitate any changes taking place in .exitlib/ as much as possible and got this solution to work without killing every session for the user but only the one targeted by the SAML logout

Best regards,
Amaury from CBlue.

Issue #417, opened 2020-04-20
Add checks when the certificate is about to expire, eg 1 month ahead


brendanheywood:
* [x] implement Check API for 3.9+
* [ ] show an warning on the admin settings page
* [ ] maybe send an email?

PR #617, opened 2021-11-15
Add certificate expiry check #417

brendanheywood:
#417


classes/check/certificateexpiry.php
```diff
@@ -0,0 +1,95 @@
...
+
+        if (time() > $expires - WEEKSECS) {
+            $summary = get_string('checkcertificatewarn', 'auth_saml2', $formatdelta);
+            return new result(result::ERROR, $summary, '');
+        }
+
+        $summary = get_string('checkcertificatewarn', 'auth_saml2', $formatdelta);
+        if (time() > $expires - 4 * WEEKSECS) {
+            return new result(result::WARNING, $summary, '');
+        }
```
keevan, line 90: I'd personally rather use something like `$now = time()` instead of getting the current time over and over.

Also a comment before each block explaining what it's doing plainly might be easier to check the logic against. Such as "Check if current time within 4hrs of expiry" (even though the check only checks for after, because the top guard should already cover that case). I think it's more understandable that way.

classes/check/certificateexpiry.php
```diff
@@ -0,0 +1,95 @@
...
+            $summary = get_string('checkcertificateexpired', 'auth_saml2', $formatdelta);
+            return new result(result::CRITICAL, $summary, '');
+        }
+
+        if (time() > $expires - WEEKSECS) {
+            $summary = get_string('checkcertificatewarn', 'auth_saml2', $formatdelta);
+            return new result(result::ERROR, $summary, '');
+        }
+
+        $summary = get_string('checkcertificatewarn', 'auth_saml2', $formatdelta);
```
keevan, line 87: This is used for both OK and WARN results. I noticed that the two strings are identical. However, it would be better to use either:
- one string for both and remove the duplicate unused one in `lang/*`,
- or just reference both separately which means if the string ever changes, it won't impact the other status type

lang/en/auth_saml2.php
```diff
@@ -47,6 +47,10 @@
 $string['attrsimple_help'] = 'Various IdP\'s such as ADFS use long attribute keys such as urns or namespaced xml schema names. If set to Yes this will simplify these, eg map http://schemas.xmlsoap.org/ws/2005/05/identity/claims/givenname to such \'givenname\'.';
 $string['certificatedetails'] = 'Certificate details';
 $string['certificatedetailshelp'] = '<h1>SAML2 auto generated public certificate contents</h1><p>The path for the cert is here:</p>';
+$string['checkcertificateexpiry'] = 'SAML certificate expiry';
+$string['checkcertificateexpired'] = 'SAML certificate has expired {$a} ago';
+$string['checkcertificatewarn'] = 'SAML certificate will expire in {$a}';
+$string['checkcertificateok'] = 'SAML certificate will expire in {$a}';
```
keevan, line 53: Referenced in earlier feedback https://github.com/catalyst/moodle-auth_saml2/pull/617/files#r748953445 but linking/mentioning for completeness

classes/check/certificateexpiry.php
```diff
@@ -0,0 +1,100 @@
...
+        $delta = time() - $expires;
+        $formatdelta = format_time($delta);
+
+        // Critical when the certicate has expired.
+        if ($now > $expires) {
+            $summary = get_string('checkcertificateexpired', 'auth_saml2', $formatdelta);
+            return new result(result::CRITICAL, $summary, '');
+        }
+
+        // Error if the certificate expiration is imminentin the next week.
```
keevan, line 84: Small typo `imminentin` should be `imminent in`

lib.php
```diff
@@ -66,3 +66,21 @@ function auth_saml2_after_require_login() {
...
+/**
+ * Add service status checks
+ *
+ * @return array of check objects
+ */
+function auth_saml2_status_checks() : array {
+    global $saml2auth;
+    require_once(__DIR__ . '/setup.php');
+
+    // If saml is not configured then return no check.
```
keevan, line 79: Perhaps it is better if the comment is be updated to reflect the immediate logic.

"If configured, it will return any status checks for the saml service" or similar.

keevan (approved):
Looks good to me now.

Issue #616, opened 2021-11-12
including settings.php file executes huge amounts of code


timhunt:
**What happened?**

On my Behat site, where I have no interest in using this plugin, I get this when going to admin notifications:

Exception - Warning: openssl_csr_sign(): cannot get CSR from parameter 1 in [dirroot]\auth\saml2\setuplib.php on line 79

More information about this error
Debug info:
Error code: generalexceptionmessage
Stack trace:

    line 157 of \lib\behat\lib.php: Exception thrown
    line ? of unknownfile: call to behat_error_handler()
    line 79 of \auth\saml2\setuplib.php: call to openssl_csr_sign()
    line 64 of \auth\saml2\setup.php: call to create_certificates()
    line 1062 of \auth\saml2\auth.php: call to require_once()
    line 298 of \auth\saml2\settings.php: call to auth_plugin_saml2->get_ssp_version()
    line 85 of \lib\classes\plugininfo\auth.php: call to include()
    line 134 of \admin\settings\plugins.php: call to core\plugininfo\auth->load_settings()
    line 8857 of \lib\adminlib.php: call to require()
    line 821 of \admin\index.php: call to admin_get_root()

**What you expected:**

1) If the plugin is being using, then it should not cause fatal erros.

2) Building the admin tree should not execute lots of code. If admin settings need fancy initialisation, then that should be lazy-loaded. This is a performance bug.

timhunt:
In my case, it is dying because of these errors (which are not reported unless I stick in print_objects) I am not going to make a separate issue for this, but 

3) create_certificates needs proper error handling. 

for me the call to `$privkey = openssl_pkey_new($opensslargs);` is generating these errors:

<ul>
<li>error:02001003:system library:fopen:No such process</li>
<li>error:2006D080:BIO routines:BIO_new_file:no such file</li>
<li>error:0E064002:configuration file routines:CONF_load:system lib</li>
<li>error:02001003:system library:fopen:No such process</li>
<li>error:2006D080:BIO routines:BIO_new_file:no such file</li>
<li>error:0E064002:configuration file routines:CONF_load:system lib</li>
</ul>

No idea what that means or how to get around it :-(

Possibly that suggests

4) If this plugin needs stuff in order to work, implement a proper environment check https://docs.moodle.org/dev/Environment_checking

brendanheywood:
+1 for the admin tree performance bug, I'll add that to the queue

[...]

PR #620, opened 2021-11-23
Issue#616: Fix performance issue caused by including settings.php file.

katerynadegtyariova:
Building admin tree should not execute the code which performs the full initialization.
For better performance the function get_ssp_version() only makes sure \SimpleSAML\Configuration is accessible through autoload.

#616

### DIFF
--- a/classes/auth.php
+++ b/classes/auth.php
@@ -1085,7 +1085,11 @@ class auth extends \auth_plugin_base {
      * @return string
      */
     public function get_ssp_version() {
-        require_once(__DIR__.'/../setup.php');
+        // To get the version there is no need to create key files and
+        // perform the full initialization. For better performance
+        // we only make sure \SimpleSAML\Configuration is accessible
+        // through _autoload.php.
+        require_once(__DIR__ . '/../_autoload.php');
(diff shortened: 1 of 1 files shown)


timhunt:
That is not the fix I would expect to see.

The point I was trying to make was, in settings.php, all the code (essentially) should be inside the `if ($ADMIN->fulltree)`.

katerynadegtyariova:
> That is not the fix I would expect to see.
> 
> The point I was trying to make was, in settings.php, all the code (essentially) should be inside the `if ($ADMIN->fulltree)`.

All the essential code in settings.php is already inside `if ($ADMIN->fulltree)` block:
https://github.com/catalyst/moodle-auth_saml2/blob/MOODLE_39_STABLE/settings.php

The problem was that the call $authplugin->get_ssp_version() run code performing the full initialization, which is not needed to get the version info:
https://github.com/catalyst/moodle-auth_saml2/blob/cd10cfd777c9c99ba3258ab9e852560e0f1195cc/settings.php#L293

timhunt:
I'm sorry. I was commenting on my memory of what the code was, rather than what is actually there. I apologies for wasting your time.

PR #621, opened 2021-11-23
Removing master branch from MOODLE_39_STABLE CI

katerynadegtyariova:
CI tests should not be running against master.


danmarsden:
I'm not convinced - why can't we run them against master? - helps to identify early issues with the branch that might require a new branch?

katerynadegtyariova:
This PR is wrong. Closing it.

PR #628, opened 2021-12-14
Add a case insensitive mapping option (3.5 back port)

doctorlard:
Initially for client usage on WR 370283

### DIFF
--- a/classes/admin/saml2_settings.php
+++ b/classes/admin/saml2_settings.php
@@ -51,4 +51,10 @@ abstract class saml2_settings {
     const OPTION_AUTO_LOGIN_SESSION = 1;
 
     const OPTION_AUTO_LOGIN_COOKIE = 2;
+
+    const OPTION_TOLOWER_EXACT = 0;
+
+    const OPTION_TOLOWER_LOWER_CASE = 1;
+
+    const OPTION_TOLOWER_CASE_INSENSITIVE = 2;
 }
--- a/classes/auth.php
+++ b/classes/auth.php
@@ -63,7 +63,7 @@ class auth extends \auth_plugin_base {
         'anyauth'            => 1,
         'idpattr'            => 'uid',
         'mdlattr'            => 'username',
-        'tolower'            => 0,
+        'tolower'            => saml2_settings::OPTION_TOLOWER_EXACT,
         'autocreate'         => 0,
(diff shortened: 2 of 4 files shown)


danmarsden:
Thanks Jonathan - rather than adding a new setting for this, I'd be more inclined to modify the existing tolower yes no select to have 3 options:
"No" = 0
"Set IDP value to lower case" = 1 (existing behaviour)
"Full case insenstive search" = 2 (new behaviour)
Along with some improved help text on that setting that covers what the different options really do... (and not using magic numbers but php constants for the 0/1/2 values...

Don't use those setting names verbatim though - I think there's probably a better way to name/describe them

Also - can you please do a PR for our MOODLE_39_STABLE branch "first" and then a MOODLE_35_STABLE backport of that patch? - thanks!

doctorlard:
Hi @danmarsden - made a 3.9 PR over at #630

doctorlard:
This patch is now deployed to client's testing Totara 12 for them to test against their Azure AD in the next day or two. Have implemented @danmarsden 's suggested changes on #630 and back-ported them (as far as possible) to this patch for 3.5 and Totara.

danmarsden:
@dmitriim do you think we could backport this one without tests? - or should we try to backport all the other test files that landed in master so we can apply here as well?

doctorlard:
Patch works on client testing Totara site. @dmitriim - it would be tricky to backport the unit test as it's embedded in `saml_login_complete` rather than its separate `user_extract` class, so I'd have to backport half the test suite from the 39 branch as well...?

dmitriim:
@doctorlard @danmarsden Basically it's not a backport, but re-implementing the patch for 3.5. I would suggest to add tests to cover new logic. It shouldn't take too much time now, but may save a lot of time later.

doctorlard:
Cool but aint nobody got time for backporting 50 tests. By "nobody" I mean the client.

danmarsden:
FYI - I looked at backporting a large chunk of stuff to the 3.5 stable branch, but I've decided we should instead just add support for T12 and higher to our 39_STABLE branch - see #644 for more details - when that lands we should be able to close this one.

danmarsden:
closing this one as won't fix as we are very close to having totara support in the 39_STABLE branch - other patches incoming soon.

PR #630, opened 2021-12-14
Add a case insensitive mapping option

doctorlard:
Initially for client usage on WR 370283; starting with `MOODLE_39_STABLE` and backport later.

### DIFF
--- a/classes/user_extractor.php
+++ b/classes/user_extractor.php
@@ -42,10 +42,11 @@ class user_extractor {
      *
      * @param string $fieldname Field name to search by.
      * @param string $fieldvalue Field value to search by.
+     * @param bool $insensitive Whether to use case insensitive match.
      *
      * @return mixed False, or A {@link $USER} object.
      */
-    public static function get_user(string $fieldname, string $fieldvalue) {
+    public static function get_user(string $fieldname, string $fieldvalue, bool $insensitive = false) {
         global $DB, $CFG;
 
         if (user_fields::is_custom_profile_field($fieldname)) {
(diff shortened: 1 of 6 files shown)


danmarsden:
Thanks @doctorlard - lgtm, @dmitriim or @brendanheywood do you guys want to +1 this before we merge it?

dmitriim:
@danmarsden Thanks for pinging me. 
I'd like to see some tests (if possible) to cover this feature. Also looks like CI is not happy (I haven't checked why).

danmarsden:
CI is complaining about behat tests for Master (moodle 4.0) - unrelated to this patch.

danmarsden:
@doctorlard - it should be pretty easy to drop in a test to the user_extractor class for your new case param which would be good to add in:
https://github.com/catalyst/moodle-auth_saml2/blob/MOODLE_39_STABLE/tests/user_extractor_test.php

I think it's probably ok to ignore the Moodle 4.0 behat issues from CI for now - we'll need to deal with them separately anyway - I'll create an issue.

classes/user_extractor.php
```diff
@@ -55,7 +56,8 @@ public static function get_user(string $fieldname, string $fieldvalue) {
 
             $joins = " LEFT JOIN {user_info_field} f ON f.shortname = :fieldname ";
             $joins .= " LEFT JOIN {user_info_data} d ON d.fieldid = f.id AND d.userid = u.id ";
-            $fieldsql = " AND d.data = :fieldvalue";
+            $datafield = $insensitive ? "LOWER(d.data)" : "d.data";
```
brendanheywood: If you are comparing case insensitive wouldn't you want to lower() both sides of the comparison?

doctorlard: We're already ensuring that `$fieldvalue` is lower case from the caller (I suppose it's only a few extra cycles and molecules of CO₂ to check in SQL too). Thoughts?

doctorlard: Actually, yeah let's fix it in SQL...

doctorlard:
> 

If it isn't a silly question, why is the CI failing Moodle 4.0 tests on a PR for `MOODLE_39_STABLE`?

brendanheywood:
I'm happy with this. We can ignore 4.0 failures

danmarsden:
> > 
> 
> If it isn't a silly question, why is the CI failing Moodle 4.0 tests on a PR for `MOODLE_39_STABLE`?

We only generate new branches when we have to fork to support a newer release. 39_Stable supports 3.9 - 3.11 and by the look of those behat errors we might have to create a new 40 branch to support 4.0 and higher. You'll find this strategy pretty common across our plugins and the readme should cover the list of branches and which moodle versions they support.

doctorlard:
> I'd like to see some tests (if possible) to cover this feature. Also looks like CI is not happy (I haven't checked why).

Test now included

dmitriim:
Thanks @doctorlard! Tests make me happy. + 1 from me for this patch.

PR #638, opened 2022-01-06
Fix code checker

tuanngocnguyen:
This patch fix issue with code checker in MOODLE 35 STABLE BRANCH

https://github.com/catalyst/moodle-auth_saml2/runs/4517404152?check_suite_focus=true


jwalits (approved):
Changes look good to me

PR #639, opened 2022-01-06
Remove reference to master branch

tuanngocnguyen:


README.md
```diff
@@ -1,5 +1,5 @@
 <a href="https://travis-ci.org/catalyst/moodle-auth_saml2">
-<img src="https://travis-ci.org/catalyst/moodle-auth_saml2.svg?branch=master">
```
keevan, line 2: I think this should be a github badge similar to changes made here?
https://github.com/catalyst/moodle-auth_saml2/pull/636/files

keevan (commented):
One small thing otherwise looks good

README.md
tuanngocnguyen: We don't have the github workflow for 27_34 So I still use the travis-ci. one

keevan: Ah okay. All good then.

PR #641, opened 2022-01-18
Moodle 35 stable simplesaml 1.19.1

danmarsden:
Top level - now that I've done the work, I'm not sure if we should merge this.. read below for more info.

Moodle 3.5-3.8 are no longer officially supported by Moodle, but the 35 stable branch is documented as the supported branch for T12 and higher so I spent some time backporting a large chunk of stuff to the 3.5 branch because we were hitting some issues running it on T15 and hitting some bugs with T12 for things that were fixed int he MOODLE_39_STABLE branch, but not the MOODLE_35_STABLE Branch..

But... now that I've mostly completed this I can see the only real differences between the 35_STABLE and 39_STABLE branches are the following:
1 privacy api class no longer using polyfill.
2 39 branch includes some workplace hooks.
3 some minor changes to test files mainly around compatibility with PHP unit 9.5 (assertEqualsCanonicalizing and assertStringContainsString) 

1 & 2 won't matter for Totara (it should ignore the privacy class and the workplace hooks will also be ignored.) 

That leaves 3) - I think we should be able to improve the tests so they will run in Totara without too many changes.

Rather than merge this in, I think we should look to moving our Totara sites to using the MOODLE_39_STABLE branch and see which remaining tests are still failing and see if we can improve them to work in both Moodle and Totara.


danmarsden:
yeah... here's an alternative to this which I think would allow unit tests on T12 and higher to pass when using the 39_STABLE branch (adding support for Totara t12 and higher to the 39 stable branch)

https://github.com/catalyst/moodle-auth_saml2/compare/MOODLE_39_STABLE-totarasupport

danmarsden:
yeah - I'm going to close this I think and instead, add support for T12 and above to our MOODLE_39_STABLE branch.

Issue #644, opened 2022-01-25
Add support for T12 and above to Moodle 3.9 branch


danmarsden:
Follow up from #641 
We created a new branch for MOODLE 39 in this repo to allow for PHP 5 to continue to be supported on an older branch - but since then, we've dropped PHP 5 support from the 3.5 stable branch.

T12 and higher seems to work ok using the MOODLE_39_STABLE branch but we hit a few different Unit test errors due to the different versions of PHP unit that run in older releases, and also due to some Totara specific requirements for Unit tests.

1) Issues around assertEqualsCanonicalizing and assertStringContainsString backwards compatibility (see #645)

2) Autoloading files needs to be more specific (see #645) - https://github.com/catalyst/moodle-auth_saml2/commit/8ae2669d38bbc7aca3572b88321686152d160dde

3) Tear down method required (see #578)

4) Differences in Generator class loading in Totara throwing errors (see #645) - example:  
```
auth_saml2\auth_test::test_is_configured
coding_exception: Coding error detected, it must be fixed by a programmer: Component auth_saml2 does not support generators. Class auth_saml2\testing\generator is missing.

/var/www/html/server/lib/classes/testing/generator.php:155
/var/www/html/server/auth/saml2/tests/auth_test.php:76
```

5) Differences in exception handling in user field tests
```
There were 2 errors:

1) auth_saml2_user_extractor_test::test_get_user_by_field_from_user_table
dml_read_exception: Error reading from database (ERROR:  column "notexists" does not exist
LINE 1: SELECT * FROM t_00user WHERE notexists = $1 AND deleted <> 1
                                     ^
SELECT * FROM t_00user WHERE notexists = $1 AND deleted <> 1
[array (
  0 => 'username2',
)])

/var/www/html/server/auth/saml2/classes/user_extractor.php:80
/var/www/html/server/auth/saml2/tests/user_extractor_test.php:91

2) auth_saml2_user_extractor_test::test_get_user_by_custom_profile_field
dml_read_exception: Error reading from database (ERROR:  column "field1" does not exist
LINE 1: SELECT * FROM t_00user WHERE field1 = $1 AND deleted <> 1
                                     ^
SELECT * FROM t_00user WHERE field1 = $1 AND deleted <> 1
[array (
  0 => 'User 1 Field 1',
)])

/var/www/html/server/auth/saml2/classes/user_extractor.php:80
/var/www/html/server/auth/saml2/tests/user_extractor_test.php:153
```

[...]

Issue #578, opened 2021-06-21
Reset test class properties after tests.


andrewmadden:
Unit test failure on MOODLE_35_STABLE branch for totara 12. 

$prophet property in `auth_saml2_metadata_fetcher_testcase` and `auth_saml2_metadata_refresh_testcase` is not reset after tests are run. Need to add tearDown method to each testcase.

```
root@a8f837e2b02e:/siteroot# vendor/bin/phpunit --testsuite auth_saml2_testsuite
Totara 12.31 (Build: 20210519.00), db92f4ef0b20e050bd546ef630db01061802493f
PHP: 7.2.24.0.0.18.04.7, pgsql: 11.4 (Debian 11.4-1.pgdg90+1), OS: Linux 5.4.0-74-generic x86_64
PHPUnit 7.5.20 by Sebastian Bergmann and contributors.

..............................................FFFFF..FFFSSSSS.... 65 / 85 ( 76%)
SSSSSSS.......S.....                                              85 / 85 (100%)

Time: 13.67 seconds, Memory: 126.50 MB

There were 8 failures:

1) auth_saml2_metadata_fetcher_testcase::test_fetch_metadata_404
Property 'prophet' defined in 'auth_saml2_metadata_fetcher_testcase' was not reset after the test!
Please either find a way to avoid using a class variable or make sure it get's unset in the tearDown method to avoid creating memory leaks.

/siteroot/lib/phpunit/classes/advanced_testcase.php:527
/siteroot/lib/phpunit/classes/advanced_testcase.php:105

To re-run:
 vendor/bin/phpunit auth_saml2_metadata_fetcher_testcase auth/saml2/tests/metadata_fetcher_test.php

2) auth_saml2_metadata_fetcher_testcase::test_fetch_metadata_success
Property 'prophet' defined in 'auth_saml2_metadata_fetcher_testcase' was not reset after the test!
Please either find a way to avoid using a class variable or make sure it get's unset in the tearDown method to avoid creating memory leaks.

/siteroot/lib/phpunit/classes/advanced_testcase.php:527
/siteroot/lib/phpunit/classes/advanced_testcase.php:105

To re-run:
 vendor/bin/phpunit auth_saml2_metadata_fetcher_testcase auth/saml2/tests/metadata_fetcher_test.php

3) auth_saml2_metadata_fetcher_testcase::test_fetch_metadata_curlerrorno
[...]

PR #645, opened 2022-01-25
T12support part 1

danmarsden:
Relates to issue #644 - fixes some unit test issues when using the 3.9 branch on T12 and higher.

This PR takes care of assertStringContainsString backwards compatiblity (using the same method as some of the other existing saml testing code), fixes a unit test failure where the autoloading classes paths are not specific enough (totara has that extra directory for wwwroot code) and ports #578 from the 3.5 stable branch into the 3.9 stable branch - it doesn't appear to affect 3.9 and higher tests.

### DIFF
--- a/tests/auth_test.php
+++ b/tests/auth_test.php
@@ -375,7 +375,11 @@ class auth_test extends \advanced_testcase {
             $this->fail('Exception expected');
         } catch (\coding_exception $e) {
             // Validate reason.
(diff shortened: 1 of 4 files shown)


keevan (approved):
Changes look good to me, I've checked both the method checks for backwards compatibility and the tearDown after test for `$this->prophet ` looks to be correct.

PR #646, opened 2022-01-26
Add testing/generator class compatible with Totara

aneno-m-e:
Added a Totara compatible generator class and edited auth_test.php and generator_test,php to handle Moodle/Totara generators.

### DIFF
--- a/classes/testing/generator.php
+++ b/classes/testing/generator.php
@@ -1,5 +1,5 @@
 <?php
-// This file is part of SAML2 Authentication Plugin
+// This file is part of Moodle - http://moodle.org/
 //
 // Moodle is free software: you can redistribute it and/or modify
 // it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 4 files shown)


danmarsden:
thanks @aneno-m-e - just summarising our internal chat:
1) lets try to find a way that doesn't involve duplicating the create_idp_entity function - either by making The totara class call the Moodle version of it or putting the function somewhere that can be re-used.
2) lets improve the phpdocs/php comments of new classes and any in-line changes to make it clear why it's being added - eg adding a comment saying "This is for Totara support"
3) if we do any class checks, lets look for the totara class to see if it exists, rather than a saml2 class - so check if \core\testing\component_generator exists rather than the saml2 class based on that class.

thanks!

danmarsden:
Thanks @aneno-m-e 

@dmitriim - I like the idea of using a trait here so that we don't duplicate the method for both Totara and Moodle - but we don't do this sort of thing very often so keen to get someone elses thoughts on this as well - any thoughts?

classes/testing/generator.php
```diff
@@ -0,0 +1,41 @@
...
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+/**
+ * @package   auth_saml2
```
dmitriim, line 27: I don't think we require this block anymore having that the class is properly documented.

classes/testing/generator.php
```diff
@@ -0,0 +1,41 @@
...
+ * @author    Noemie Ariste <noemie.ariste@catalyst.net.nz>
+ * @copyright Catalyst IT
+ * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+
+namespace auth_saml2\testing;
+
+use stdClass, coding_exception;
```
dmitriim: It's not a common way for moodle to format your use. Usually each use is on a new line.

classes/testing/tests_generator.php
```diff
@@ -0,0 +1,78 @@
...
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+/**
+ * @package   auth_saml2
```
dmitriim: Same here. Should be moved to a class doc.

dmitriim:
Thanks @danmarsden  for pinging me. I think more common way of doing this is having a base generator class with common methods and then child generators can extent this base class. However, I don't mind to use trait in this case. I think I have done the same for one of my plugins. 

+1 from me.

PR #655, opened 2022-02-15
Add check for Totara 12 compatibility

aneno-m-e:
If function doesn't exist, skip calls to component_class_callback so unit tests can pass with T12.

### DIFF
--- a/classes/auth.php
+++ b/classes/auth.php
@@ -224,8 +224,9 @@ class auth extends \auth_plugin_base {
             }
 
             // Moodle Workplace - Check IdP's tenant availability.
-            if (!component_class_callback('\tool_tenant\local\auth\saml2\manager', 'issuer_available',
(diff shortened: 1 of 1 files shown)


danmarsden:
@dmitriim I'm on the fence if this reads well enough or if we should change this to a class_exists call a bit like:
```
if (class_exists('\tool_tenant\local\auth\saml2\manager'))
```
instead of the check to see if component_class_callback() function exists - the issue is that component_class_callback() doesn't exist in T12 (it does in later versions of Totara though.)

what do you think?

dmitriim:
up to you really, I don't have strong preferences here if it works the same way

From internal chat: 

Yet another alternative:
class_exists(\tool_tenant\local\auth\saml2\manager::class) which your ide might pick up

danmarsden:
@kabalin - FYI - you might want to be aware of this change before it lands.

Issue #661, opened 2022-02-28
WantAssertionsSigned property


Tassleh0ff:
I found this pull request that added WantAssertionsSigned property to the plugin.
https://github.com/catalyst/moodle-auth_saml2/pull/585

I noticed the plugin I installed from Moodle was a version prior to the merge of this new feature into the main branch, so I updated throught git and I got the new option "Want assertions signed", but if I change it from default value (No) to Yes and save the configuration, my metadata does not change and there's no WantAssertionsSigned field in the xml.

Do I need to change some other options or add something in the config.php or any other file?
Thanks for any information :)

alexmorrisnz:
Hi @Tassleh0ff can you regenerate the metadata and see if it has updated with the field?
Visiting the url `/auth/saml2/sp/metadata.php?regenerate=1` will regenerate the file

Tassleh0ff:
Worked perfectly, now the metadata contains the WantAssertionsSigned parameter.
Thanks!

PR #662, opened 2022-03-01
Add sp metadata update callback to more settings

alexmorrisnz:
This adds the  `auth_saml2_update_sp_metadata` callback to settings that require a change in the SP metadata.

`auth_saml2_update_sp_metadata` was missing the call to `setup.php` which is required to initialise `$saml2auth`.

It will prevent problems such as #661


alexmorrisnz:
Looks like the behat failures are an existing issue and not caused by this PR

danmarsden:
rebased in  branch rebase662

PR #664, opened 2022-03-03
Rename entityid to spentityid

alexmorrisnz:


lang/en/auth_saml2.php
```diff
@@ -116,8 +116,8 @@
 $string['wantassertionssigned_help'] = 'Whether assertions received by this SP must be signed';
 $string['assertionsconsumerservices'] = 'Assertions consumer services';
 $string['assertionsconsumerservices_help'] = 'List of bindings the SP should support';
-$string['entityid'] = 'Entity ID';
-$string['entityid_help'] = 'Unique name of the entity provider. Leave blank unless required';
+$string['spentityid'] = 'Entity ID';
+$string['spentityid_help'] = 'Unique name of the entity provider. Leave blank unless required';
```
brendanheywood: actually isn't this string wrong too? should this be the 

Entity Id of the Service Provider ...

alexmorrisnz: yep will change it

brendanheywood: I would also be a bit stronger with the lang, like 

Override the Entity Id of the Service Provider. In most cases leave blank and a good default will be used instead.

alexmorrisnz: I've set it to that

PR #665, opened 2022-03-03
Add upgrade step for deleting entityid setting

alexmorrisnz:


db/upgrade.php
```diff
@@ -368,5 +368,13 @@ function xmldb_auth_saml2_upgrade($oldversion) {
         upgrade_plugin_savepoint(true, 2020082100, 'auth', 'saml2');
     }
 
+    if ($oldversion < 2022030300) {
+        // Delete old configuration value that is no longer used, to prevent any future conflicts.
+        set_config('entityid', null, 'auth_saml2');
```
brendanheywood: this is fine, but unset_config would be better

alexmorrisnz: Updated, thanks
Didn't know that was a function, hadn't scrolled down far enough in moodlelib to see it.

PR #667, opened 2022-03-07
Re-Add check for Totara 12 compatibility

aneno-m-e:
If function doesn't exist, skip calls to component_class_callback so unit tests can pass with T12.

This code was submitted in this PR https://github.com/catalyst/moodle-auth_saml2/pull/655 and merged but is now missing from the codebase. If absent, it breaks phpunit tests on Totara 12.


danmarsden:
thanks - probably my bad - PR #655 was sent to the wrong branch and not to MOODLE_39_STABLE so probably a manual merge I did didn't catch it all - thanks for the PR, when tests pass I'll merge.

PR #673, opened 2022-03-15
Fix behat and unit tests

danmarsden:
@dmitriim do you want to review the changes to teh user_extrator class to make it case sensitive for Mariadb based sites?

also - do we need to be aware of this changing existing behaviour for mariadb based sites who will have been getting case insensitive results even though the default setting for sensistivity is set to an exact match?


danmarsden:
I've just added an extra commit to set the case senstivity to "insensitve" for existing sites that are mysql/mariadb based - this will be how they are operating at the moment with the old code, so hopefully this prevents a potential issue with mariadb sites when we merge this so that it respects the case sensitivity setting.

brendanheywood:
+1 from code review from me, waiting for ci

dmitriim (approved):

PR #675, opened 2022-03-16
Add supported versions to version.php

danmarsden:


danmarsden:
will wait for tests to pass, but then merge to trigger the plugins db update.

Issue #679, opened 2022-03-30
case sensitivity only works for custom fields


danmarsden:
The new case insensitivity setting only seems to work with custom fields - not core user fields.

PR incoming.

PR #681, opened 2022-04-04
Handle case sensitivity for core user fields

aneno-m-e:
Fix for issue #679 

Update class user_extractor.
Add new tests cases to class with core user fields.

### DIFF
--- a/classes/user_extractor.php
+++ b/classes/user_extractor.php
@@ -48,6 +48,9 @@ class user_extractor {
         global $DB, $CFG;
 
         $user = false;
+        $joins = '';
+        $params['fieldvalue'] = $fieldvalue;
+        $params['mnethostid'] = $CFG->mnet_localhost_id;
 
         if (user_fields::is_custom_profile_field($fieldname)) {
 
@@ -58,13 +61,21 @@ class user_extractor {
 
             $fieldsql = " AND ". $DB->sql_equal('d.data', ':fieldvalue', !$insensitive);
             $params['fieldname'] = $fieldname;
-            $params['fieldvalue'] = $fieldvalue;
-            $params['mnethostid'] = $CFG->mnet_localhost_id;
 
+        } else {
+            // Check if requested field exists, required for Totara compatibility.
+            $fields = array_merge(\core_user::AUTHSYNCFIELDS, ['id', 'username']);
+            if (in_array($fieldname, $fields)) {
+                $fieldsql = " AND ". $DB->sql_equal('u.' . $fieldname, ':fieldvalue', !$insensitive);
+                $params['fieldname'] = $fieldname;
+            }
+        }
+
+        if (!empty($fieldsql)) {
             $sql = "SELECT u.id
                       FROM {user} u $joins
-                     WHERE u.deleted <> 1 AND
-                           u.mnethostid = :mnethostid $fieldsql";
+                     WHERE u.deleted <> 1
+                       AND u.mnethostid = :mnethostid $fieldsql";
 
             if ($records = $DB->get_records_sql($sql, $params)) {
                 if (count($records) == 1) {
@@ -72,15 +83,7 @@ class user_extractor {
                     $user = get_complete_user_data('id', $record->id);
                 }
             }
-        } else {
-            // Check if requested field exists, required for Totara compatibility.
-            $fields = array_merge(\core_user::AUTHSYNCFIELDS, ['id', 'username']);
-            if (in_array($fieldname, $fields)) {
-                $user = get_complete_user_data($fieldname, $fieldvalue);
-            }
         }
-
         return $user;
     }
-
 }
--- a/tests/user_extractor_test.php
+++ b/tests/user_extractor_test.php
@@ -153,13 +153,13 @@ class auth_saml2_user_extractor_test extends advanced_testcase {
     }
 
     /**
-     * Tests for case insensitive match.
+     * Tests for case insensitive match on custom user fields.
      */
-    public function test_get_user_case_insensitive() {
+    public function test_get_user_case_insensitive_custom_fields() {
         global $CFG;
         $this->resetAfterTest();
 
-        // Arrange data
+        // Arrange data.
         $field = $this->add_user_profile_field('vehicleplate', 'text', true);
         $expecteduser = $this->getDataGenerator()->create_user();
         profile_save_data((object)['id' => $expecteduser->id, 'profile_field_vehicleplate' => 'HD4999']);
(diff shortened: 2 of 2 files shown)


tests/user_extractor_test.php
```diff
@@ -159,11 +159,39 @@ public function test_get_user_case_insensitive() {
...
+        // Should match with different case.
+        $actualuser = user_extractor::get_user('idnumber', 'nb256', true);
+        $this->assertNotFalse($actualuser);
+        $this->assertSame($expecteduser->id, $actualuser->id);
+
+        // Should not match with different value (obviously).
+        $actualuser = user_extractor::get_user('idnumber', 'Some other value entirely', true);
+        $this->assertFalse($actualuser);
+
+        if (!in_array($CFG->dbtype, ['mysqli', 'mariadb'])) {
```
danmarsden: @dmitriim thoughts on this in the tests? - I'm ok with it and there's no easy API that says "does your db type support case sensitivity"

keevan: There is `$DB->get_dbfamily()` that might be useful here. Perhaps this affects the mysql family. (instead of needing to list out each one)

https://github.com/moodle/moodle/blob/6f58f72/lib/accesslib.php#L5289

tests/user_extractor_test.php
```diff
@@ -159,11 +159,39 @@ public function test_get_user_case_insensitive() {
...
         $this->resetAfterTest();
 
-        // Arrange data
+        // Arrange data.
         $field = $this->add_user_profile_field('vehicleplate', 'text', true);
-        $expecteduser = $this->getDataGenerator()->create_user();
+        $expecteduser = $this->getDataGenerator()->create_user(['idnumber' => 'NB256']);
         profile_save_data((object)['id' => $expecteduser->id, 'profile_field_vehicleplate' => 'HD4999']);
 
+        // First, test get_user using a core field.
```
dmitriim: I'd suggest to split this into two tests: one for profile fields and one for core fields.

classes/user_extractor.php
```diff
@@ -76,7 +76,26 @@ public static function get_user(string $fieldname, string $fieldvalue, bool $ins
...
             $fields = array_merge(\core_user::AUTHSYNCFIELDS, ['id', 'username']);
             if (in_array($fieldname, $fields)) {
-                $user = get_complete_user_data($fieldname, $fieldvalue);
+                if ($insensitive) {
+                    $fieldsql = " AND ". $DB->sql_equal($fieldname, ':fieldvalue', !$insensitive);
+                    $params['fieldname'] = $fieldname;
+                    $params['fieldvalue'] = $fieldvalue;
+                    $params['mnethostid'] = $CFG->mnet_localhost_id;
+
+                    $sql = "SELECT id
```
dmitriim: If we use a custom SQL in some cases, would it be better to use it all the time? Would probably require a little bit of refactoring to remove a duplicating code.

tests/user_extractor_test.php
aneno-m-e: I've split the test into two. I'll wait to see where the db type discussion lands.

classes/user_extractor.php
```diff
@@ -76,7 +76,23 @@ public static function get_user(string $fieldname, string $fieldvalue, bool $ins
...
+
+                $fieldsql = " AND ". $DB->sql_equal($fieldname, ':fieldvalue', !$insensitive);
+                $params['fieldname'] = $fieldname;
+                $params['fieldvalue'] = $fieldvalue;
+                $params['mnethostid'] = $CFG->mnet_localhost_id;
+
+                $sql = "SELECT id
+                            FROM {user}
+                            WHERE deleted <> 1 AND
+                                mnethostid = :mnethostid $fieldsql";
```
keevan:
```suggestion
                            WHERE deleted <> 1 
                              AND mnethostid = :mnethostid $fieldsql";
```

classes/user_extractor.php
```diff
@@ -76,7 +76,23 @@ public static function get_user(string $fieldname, string $fieldvalue, bool $ins
...
+
+                if ($records = $DB->get_records_sql($sql, $params)) {
```
keevan: Perhaps you can use `$DB->get_record_sql($sql, $params, IGNORE_MULTIPLE)` here instead, with the `IGNORE_MULTIPLE` flag to only use the first result if one exists. This should simplify the logic in this block.

Reference: https://docs.moodle.org/dev/Data_manipulation_API#get_record_sql

danmarsden: But if more than  one record we exit right? And don't allow the match? Your example would allow a match when more than one user is possible to select from?

keevan: Ah, you're right. I missed the intention. The alternative here would be using the `MUST_EXIST` flag, however that will look like making a code flow work with exception handling which is even uglier.

In this case, the only suggest would be the use `===` instead of `==` for the count check.

classes/user_extractor.php
```diff
@@ -76,7 +76,23 @@ public static function get_user(string $fieldname, string $fieldvalue, bool $ins
             // Check if requested field exists, required for Totara compatibility.
             $fields = array_merge(\core_user::AUTHSYNCFIELDS, ['id', 'username']);
             if (in_array($fieldname, $fields)) {
-                $user = get_complete_user_data($fieldname, $fieldvalue);
+
+                $fieldsql = " AND ". $DB->sql_equal($fieldname, ':fieldvalue', !$insensitive);
```
dmitriim: There is still a lot of duplicated code here. As now we use SQL for both paths of if statement we can move actual SQL execution and user search out of it. In if else we would build SQL's parts as they are different depending on scenario. Hope that makes sense.

aneno-m-e: Hi @dmitriim I've refactored as you were suggesting. Can you have a look and hopefully merge?

dmitriim: Sorry for being a pain, but it's still have some code duplication. We basically rewrite the same SQL in both parts of if statement.

dmitriim: How about something like:

`    /**
     * Get extracted from DB user.
     *
     * @param string $fieldname Field name to search by.
     * @param string $fieldvalue Field value to search by.
     * @param bool $insensitive Whether to use case insensitive match.
     *
     * @return mixed False, or A {@link $USER} object.
     */
    public static function get_user(string $fieldname, string $fieldvalue, bool $insensitive = false) {
        global $DB, $CFG;

        $user = false;
        $joins = '';
        $params['fieldvalue'] = $fieldvalue;
        $params['mnethostid'] = $CFG->mnet_localhost_id;

        if (user_fields::is_custom_profile_field($fieldname)) {

            $fieldname = user_fields::get_field_short_name($fieldname);

            $joins .= " LEFT JOIN {user_info_field} f ON f.shortname = :fieldname ";
            $joins .= " LEFT JOIN {user_info_data} d ON d.fieldid = f.id AND d.userid = u.id ";

            $fieldsql = " AND ". $DB->sql_equal('d.data', ':fieldvalue', !$insensitive);
            $params['fieldname'] = $fieldname;

        } else {
            // Check if requested field exists, required for Totara compatibility.
            $fields = array_merge(\core_user::AUTHSYNCFIELDS, ['id', 'username']);
            if (in_array($fieldname, $fields)) {
                $fieldsql = " AND ". $DB->sql_equal('u.' . $fieldname, ':fieldvalue', !$insensitive);
                $params['fieldname'] = $fieldname;

            }
        }

        if (!empty($fieldsql)) {
            $sql = "SELECT u.id
                          FROM {user} u $joins
                         WHERE u.deleted <> 1 AND
                               u.mnethostid = :mnethostid $fieldsql";

            if ($records = $DB->get_records_sql($sql, $params)) {
                if (count($records) == 1) {
                    $record = reset($records);
                    $user = get_complete_user_data('id', $record->id);
                }
            }
        }

        return $user;
    }`

dmitriim (approved):

LiamKearn:
IMO a bit of a breaking (but needed) change. Anyone who has installed since Dec 2021 and deemed the **default** settings (exact matching with core_field `username`) are working could be caught off guard by this change causing **the default config convention** to functions completely differently and create duplicate user accounts due to case mismatches in mdlattr.

#### Constructively:
I think insensitive should have been changed to the default after this change. This doesn't account for custom profile fields though since those were working with exact matching before nor is it a  SAML ideal.

Perhaps another solution would be to use the upgrade.php to run a migration converting instances that specifically use a core profile field `mdlattr` to case insensitive matching then administrators can opt-in to exact matching rather then having it unexpectedly working when it did not prior to upgrade (I'd be happy to PR this second option if deemed reasonable).

#### Question:
Given that MOODLE version numbers are just incremental / date based is there any mechanism that is generally used to indicate breaking changes?

danmarsden:
unfortuantely it wasn't quite that simple.... by setting case insensitive as the default you end up with the possibility (although unlikely sure.) where users who should not have been able to login and weren't in some earlier versions of the code being able to login depending on which attribute was being used for matching.
While it's inconvenient for an admin to have to change this setting, the risk of it presenting a security issue with users being able to login seemed more important. 

Sorry for any inconveneince!

LiamKearn:
Ah Just found: https://github.com/catalyst/moodle-auth_saml2/commit/9486585f08f91e3be8dcbb93de562d8ab9bce4cf. I didn't realise the exact was the default in the past I assumed that it was insensitive. This makes more *sense* now. Thanks :)

PR #682, opened 2022-04-04
Allow generation of metadata sp for another domain in prep for dns move

### LINKED ISSUE #615: Have a new feature to allow a site to generate SP metadata for a different domain
The use case is a new site which is not yet ready, but will eventually be a replacement for another production site. We cannot use the real domain name yet until dns cutover.

But in theory we can generate the SP metadata and all the files in the correct places ahead of time. There should be no security issues here as its equivalent to someone manually grabbing the xml file and making their own changes, which we cannot prevent and has no downside.

We'd still only allow this to be done for people who are logged in as admin to generate the internal files in the right place, and then they'll be available for public download (similar to how regeneration is done)

Tasks:
* [ ] Have a magic url which be used to generate a new set of files on demand
* [ ] have integration with local_envar which detects if it is non prod and can generate prod ones, or vice versa if it is prod then it can generate the stage / uat ones. In both cases you can link to / download the matrix of files

Manual workaround steps:

* Go to /var/lib/sitedata/saml2/ for the pre prod site and copy the existing CRT and PEM files to the correct filename;
* change the wwwroot to the eventual prod url
* fake dns eg update laptop /etc/hosts to point the prod url
* Browse to the now non prod site which thinks it is prod
* make it generate a new XML as normal
* Check /var/lib/sitedata/ and confirm it didn't blow away the existing crt / pem files
* Download / grab the new XML file

brendanheywood:
Fixed #615

Usage is go to the normal SP page and then enter an extra query param eg:

http://current.edu/auth/saml2/sp/metadata.php?baseurl=https://newdomain.com

It can also be useful to regenerate it (must be logged in as admin) while testing to make sure it is refreshed:

http://moodle.local/auth/saml2/sp/metadata.php?regenerate=1&baseurl=https://acme.com

This will:
1) make new crt and pem files under the right file names so that after the dns move they will be in the correct place and in theory just work
2) generate the sp xml metadata in the right shape. This is half working with the sp name but not the domain name bits in the xml

Rough test steps:
1) In your idp, ingest the new SP metadata such as http://moodle.local/auth/saml2/sp/metadata.php?regenerate=1&baseurl=https://acme.com
2) Allow what will be the new SP's entityid
2) Hack hosts or otherwise simulate a DNS change and voila login should continue to work

### DIFF
--- a/classes/auth.php
+++ b/classes/auth.php
@@ -100,10 +100,13 @@ class auth extends \auth_plugin_base {
         $this->defaults['idpdefaultname'] = get_string('idpnamedefault', 'auth_saml2');
         $this->defaults['flagmessage'] = get_string('flagmessage_default', 'auth_saml2');
         $this->authtype = 'saml2';
-        $mdl = new moodle_url($CFG->wwwroot);
+
+        $baseurl = optional_param('baseurl', $CFG->wwwroot, PARAM_URL);
+        $mdl = new moodle_url($baseurl);
         $this->spname = $mdl->get_host();
         $this->certpem = $this->get_file("{$this->spname}.pem");
         $this->certcrt = $this->get_file("{$this->spname}.crt");
+
         $this->config = (object) array_merge($this->defaults, (array) get_config('auth_saml2') );
 
         // Parsed IdP metadata, either a list of IdP metadata urls or a single XML blob.
--- a/config/config.php
+++ b/config/config.php
@@ -38,11 +38,12 @@ foreach ($saml2auth->metadataentities as $idpentity) {
 }
 
 $remoteip = getremoteaddr();
+$baseurl = optional_param('baseurl', $CFG->wwwroot, PARAM_URL);
 
 $config = array(
-    'baseurlpath'       => $CFG->wwwroot . '/auth/saml2/sp/',
+    'baseurlpath'       => $baseurl . '/auth/saml2/sp/',
     'application'       => [
-      'baseURL'         => $CFG->wwwroot . '/auth/saml2/sp/',
+      'baseURL'         => $baseurl . '/auth/saml2/sp/',
     ],
     'certdir'           => $saml2auth->get_saml2_directory() . '/',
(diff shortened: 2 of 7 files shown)


config/config.php
```diff
@@ -39,10 +39,13 @@
 
 $remoteip = getremoteaddr();
 
+$baseurl = optional_param('baseurl', $CFG->wwwroot, PARAM_URL);
+
```
dmitriim, line 42: extra empty lines?

sp/metadata.php
```diff
@@ -39,13 +39,21 @@
     header('Content-Disposition: attachment; filename=' . $saml2auth->spname . '.xml');
 }
 
+if (is_siteadmin()) {
```
dmitriim: does that mean we would need to download xml as admins and won't be able to provide it as URL?

brendanheywood: I removed this so it can be public

classes/auth.php
```diff
@@ -100,10 +100,13 @@ public function __construct() {
         $this->defaults['idpdefaultname'] = get_string('idpnamedefault', 'auth_saml2');
         $this->defaults['flagmessage'] = get_string('flagmessage_default', 'auth_saml2');
         $this->authtype = 'saml2';
-        $mdl = new moodle_url($CFG->wwwroot);
+
+        $baseurl = optional_param('baseurl', $CFG->wwwroot, PARAM_URL);
```
Peterburnett, line 104: Could this have the potential to overwrite the current prod cert via using the default value?

brendanheywood: Nope, if the base != wwwroot then it's an independent file

Peterburnett: NVM this param only ever concerns the metadata XML, not the underlying certs

Peterburnett (approved):

danmarsden:
this didn't pass tests.... and introduced a new failure... whoops?

PR #686, opened 2022-05-05
Fix the settings page exception caused by trying to call a function on a NULL.

### LINKED ISSUE #656: Settings page exception on save after clean installation
This may be connected to #652, because get_file_idp_metadata_file is also affecting the end-to-end flow of clean installation of this plugin.
Using HEAD of the MOODLE_39_STABLE branch at commit #858a056f, I can install the plugin and proceed through the installation as far as the plugin settings page but saving that page without changing any default values causes an exception:
```
Exception - Call to a member function get_file_sp_metadata_file() on null
Debug info:
Stack trace:
line 284 of /auth/saml2/locallib.php: Error thrown
line 2051 of /lib/adminlib.php: call to auth_saml2_update_sp_metadata()
line 8968 of /lib/adminlib.php: call to admin_setting->post_write_settings()
line 26 of /admin/upgradesettings.php: call to admin_write_settings()
```
This seems to be because /var/moodledata/saml2 has not been populated with .cert, .pem and .xml files at this point, so auth/saml2/setup.php fails at the try/catch. Requesting an XML download populates the directory with those files, but this is not possible during the install process.
It seems to me that it should be possible to install the plugin without configuring it at the outset. If it does need to be configured at installation time, a note on the documentation page to say so would be a huge help!

cwarwicker:
Fixes: #656

### DIFF
--- a/locallib.php
+++ b/locallib.php
@@ -279,7 +279,10 @@ function auth_saml2_get_sp_metadata($baseurl = '') {
  *
  */
 function auth_saml2_update_sp_metadata() {
-    global $saml2auth;
+    global $CFG, $saml2auth;
+
+    // Require the setup file to initialise the creation of any missing cert files and/or directories.
+    require_once $CFG->dirroot . '/auth/saml2/setup.php';
 
     $file = $saml2auth->get_file_sp_metadata_file();
     @unlink($file);


danmarsden:
looks like a similar change also proposed in #662 but we weren't aware that would also fix #656

cwarwicker:
Just following up, as the client has chased. Do you think this, or the other similiar fix, are likely to be reviewed/approved soonish?

Thanks.

danmarsden:
just dropped in a rebase of Alex's patch - if tests pass, I'll merge it in, (closing this one.)

PR #695, opened 2022-05-25
Fix #694 - create user with correct username.

### LINKED ISSUE #694: When user mapping is not done on username field, initial user creation uses an invalid username.
When saml config is set up to map something like Moodle ID Number with an external value and also set to create new users, 

this line:
https://github.com/catalyst/moodle-auth_saml2/blob/MOODLE_39_STABLE/classes/auth.php#L664

Will create a new user account, but will populate the username field with the external "idnumber" value.

Then immediately after - it will replace the username with the correct value here:
https://github.com/catalyst/moodle-auth_saml2/blob/MOODLE_39_STABLE/classes/auth.php#L724

This can cause some weird edge case issues where an account already exists with the external idnumber stored as a username in  moodle (so the user creation causes a db  failure on the username index)

This is dodgy anyway - it should create the new user record and use the correct external username rather than creating a user record with an incorrect username.

danmarsden:
untested patch to create username with correct username when username is not the uid.

### DIFF
--- a/classes/auth.php
+++ b/classes/auth.php
@@ -660,8 +660,14 @@ class auth extends \auth_plugin_base {
                     $this->handle_blocked_access();
                 }
 
-                $this->log(__FUNCTION__ . " user '$uid' is not in moodle so autocreating");
-                $user = create_user_record($uid, '', 'saml2');
+                $username = $this->get_username_from_attributes($attributes);
+                if (empty($username)) {
+                    // Just in case username field not set, use uid.
+                    $username = $uid;
+                }
+
+                $this->log(__FUNCTION__ . " user '$username' is not in moodle so autocreating");
+                $user = create_user_record($username, '', 'saml2');
                 $newuser = true;
             } else {
                 // Moodle user does not exist and settings prevent creating new accounts.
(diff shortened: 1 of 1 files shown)


classes/auth.php
```diff
@@ -962,6 +968,21 @@ public function get_email_from_attributes(array $attributes) {
...
     }
 
+    /**
+     * Get lowercase username from attributes, force as lowercase because Moodle requires it.
+     *
+     * @param array $attributes A list of attributes.
+     *
+     * @return bool
+     */
+    private function get_username_from_attributes(array $attributes) {
```
Peterburnett, line 978: Probably worth adding the return type hint

classes/auth.php
```diff
@@ -962,6 +968,21 @@ public function get_email_from_attributes(array $attributes) {
...
+        if (!empty($this->config->field_map_username) && !empty($attributes[$this->config->field_map_username])) {
+            return strtolower($attributes[$this->config->field_map_username][0]);
```
Peterburnett, line 980: Could this ever be a string not an array in the SAML response?

danmarsden: good thought - I traced through the code to see where it would usually set the username value in the update_user call and found:
https://github.com/catalyst/moodle-auth_saml2/blob/MOODLE_39_STABLE/classes/auth.php#L924

Which shows it's always treated as an array in the existing code, so we can leave this for now and come back to it later if needed.

classes/auth.php
danmarsden: hmm - return type is actually mixed string|bool - so I'll update the phpdoc, but there isn't a php 7 compat way of adding in the return type mixed on the function is there?

Peterburnett: No dont think there is :(

Peterburnett (approved):

PR #700, opened 2022-06-12
When username field is updated, be consistent about how we set it.

danmarsden:
when the username is updated, we should make sure it is set to lower case in the same way we do when creating the user.

### DIFF
--- a/classes/auth.php
+++ b/classes/auth.php
@@ -925,9 +925,12 @@ class auth extends \auth_plugin_base {
                                         continue;
                                     }
                                 }
-
-                                // Custom profile fields have the prefix profile_field_ and will be saved as profile field data.
-                                $user->$field = $attributes[$attr][0];
+                                if ($field == 'username') {
+                                    $user->$field = strtolower($attributes[$attr][0]);
+                                } else {
+                                    // Custom profile fields have the prefix profile_field_ and will be saved as profile field data.
+                                    $user->$field = $attributes[$attr][0];
+                                }
                                 $update = true;
                             }
                         }
--- a/tests/auth_test.php
+++ b/tests/auth_test.php
@@ -1358,6 +1358,32 @@ class auth_test extends \advanced_testcase {
     }
 
     /**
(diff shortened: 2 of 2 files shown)


classes/auth.php
```diff
@@ -925,9 +925,12 @@ public function update_user_profile_fields(&$user, $attributes, $newuser = false
                                         continue;
                                     }
                                 }
-
-                                // Custom profile fields have the prefix profile_field_ and will be saved as profile field data.
-                                $user->$field = $attributes[$attr][0];
+                                if ($field == 'username') {
+                                    $user->$field = strtolower($attributes[$attr][0]);
```
dmitriim, line 929: do we have a test that covers that scenario?

danmarsden: yeah - there are some tests that are passing, but there's an edge case here that I don't think we have a test for - I'll try to come back and add one in.

I'm also looking closely at this block:
https://github.com/catalyst/moodle-auth_saml2/blob/MOODLE_39_STABLE/classes/auth.php#L920-L927

if the external idp value is matching on something like idnumber - it should be possible to update the moodle username IMO.

dmitriim: Yes agree. It feels like this code needs some refactoring to include all these scenarios and still be understandable.

dmitriim: So basically only matching fields should be locked from updating. This will require to add more logic for checking things + tests.

danmarsden: in the meantime - I can see a test I can add for this specific case - just running it through now so I can get this fix in and we can come back to the fact that username can't be updated later.

danmarsden:
yeah - test added now and I can see it passing.

I also pushed that same test into an old branch before this patch and you can see it failing here:
https://github.com/catalyst/moodle-auth_saml2/actions/runs/2491868632

dmitriim (approved):

classes/auth.php
dmitriim: Lets' create an issue to come back to it later. Otherwise LGTM.

PR #708, opened 2022-07-08
Allow fields with multivalued responses returned from the idp to be mapped to user profile fields (Partially resolves #346)

### LINKED ISSUE #346: MultiValued attributes
**What Happened ?**

https://github.com/catalyst/moodle-auth_saml2/blob/04044eabec6bf0841d8f11f5a0da4abed53ae883/auth.php#L676

In the referenced line of the code, if multiple values are passed by the IdP are automatically discarded because always the first returned value is taken. 

**What you expected**
Would be nice to have support for multi valued attributes coming from the IdP.

mattporritt:
Responses from the IdP can look like:

```
{
  "email": ["mattp@some.fake.domain"],
  "givenName": ["Matt"],
  "sn":["Porritt"],
  "choices": ["first","second","third"]
}
```
In the current implementation of the plugin, returned fields with multiple values only save the first value in the array into user profile fields.
This patch converts fields with many values to a comma separated string. This can then be saved in a profile field.


mattporritt:
ahhhh damn you old Moodle and your lack of test generators.
I really want to branch the plugin because of this. But that's probably overkill

classes/auth.php
```diff
@@ -929,7 +929,7 @@ public function update_user_profile_fields(&$user, $attributes, $newuser = false
                                     $user->$field = strtolower($attributes[$attr][0]);
                                 } else {
                                     // Custom profile fields have the prefix profile_field_ and will be saved as profile field data.
-                                    $user->$field = $attributes[$attr][0];
+                                    $user->$field = implode(',', (array) $attributes[$attr]);
```
keevan: Just sanity checking, if the field set in attributes[attr] was NULL, is it expected that the value stored is an empty string (as it would be with this particular change),  or NULL?

mattporritt: I think an empty string is fine. Null isn't a special case/value for these fields.

brendanheywood (approved):

Issue #569, opened 2021-06-15
.extlib/simplesamlphp/modules/.gitignore ignores directories authYubiKey and metarefresh


danowar2k:
We store our custom Moodle version in a git repository and update used 3rd party plugin versions in that repository.

We've now updated to https://moodle.org/plugins/auth_saml2/2021040100/24125 .
I've just now noticed when switching branches that there is a .gitignore file 

.extlib/simplesamlphp/modules/.gitignore which leads to our git ignoring the following directories:

* .extlib/simplesamlphp/modules/authYubiKey
* .extlib/simplesamlphp/modules/metarefresh

We deploy our Moodle from a git branch/tag. I'm now confused and wonder if those two directories should be present in our git repository and must be deployed as well or if this is a simple "We added these directories to gitignore but didn't remove them from tracking".

Please explain.

danmarsden:
@danowar2k - .extlib/simplesamlphp is a copy of the upstream code from the simplesaml project. I see the "old" .gitignore did exlude that directory, but the one currently in core doesn't - I suspect something else in your config is ignoring ".gitignore" files in your system so you've ended up getting the "old" .gitignore file rather than the current one in this release.

I'm open to the idea of removing it from the .extlib/simplesamlphp/modules directory, but we prefer to include simplesamlphp "as-is" from upstream with the minimum number of changes as this make it easier to upgrade to the newer library when it is released.

danowar2k:
I just found the release script of simplesamlphp which seems to remove the "modules/.gitignore" file in releases.

https://github.com/catalyst/moodle-auth_saml2/blob/ec90cd734517e63b0e2fde9ebf4ae33691752b75/.extlib/simplesamlphp/bin/build-release.sh#L67

So if I understand it correctly you're not dropping a released simplesamlphp in your plugin (because it would then not have the .gitignore files) but the upstream version, meaning you also clone simplesamlphp from source, at the moment probably the 1.18.8 tag. Right?

I'm gonna go look at my Moodle project's own .gitignore files, I'm currently confused because I don't remember doing any ignoring of other .gitignore files...we'll see.

Can you point me to the "old" .gitignore you mentioned? Looking myself just in case...

danowar2k:
Okay, I've looked at our .gitignore files...

We basically have
/our-repo/.gitignore (only ignores .idea stuff and specifically named directories) (If requested I could post it)
[...]

PR #710, opened 2022-07-21
#569 remove .gitignore file causing problems for git-managed deployments

davosmith:
#569


danmarsden:
Thanks @davosmith - nice catch on the broken test too! - thanks for fixing that up!

PR #715, opened 2022-08-12
Fully hydrate user account before creating it

catalystfd:
This prevents issues with plugins such as local welcome, where they
cannot function correctly as the account is created with only minimal
information such as the username.

This extracts the logic for extracting saml2 attributes out of the
update function so both update and create can use it, and then fully
hydrates a user object before calling user_create_user to create the
moodle user object.

Password updating is set to true for the use case where a different
auth plugin is being used that might support the password field.

I still need to test this against a test IDP before it's ready for final merge
but opening now for any review comments.

### DIFF
--- a/tests/auth_test.php
+++ b/tests/auth_test.php
@@ -673,12 +673,9 @@ class auth_saml2_test extends \advanced_testcase {
 
         // Checking that the events contain the expected values.
         $events = $sink->get_events();
-        $this->assertCount(4, $events);
+        $this->assertCount(2, $events);
         $event = array_pop($events);
         $this->assertInstanceOf('\core\event\user_loggedin', $event);
-        $this->assertEquals($USER->id, $event->get_data()['objectid']);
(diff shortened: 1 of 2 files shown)


tests/auth_test.php
```diff
@@ -673,14 +673,11 @@ public function test_saml_login_complete_new_account(): void {
 
         // Checking that the events contain the expected values.
         $events = $sink->get_events();
-        $this->assertCount(4, $events);
+        $this->assertCount(2, $events);
```
catalystfd, line 676: Now that user creation fully hydrates the user profile, the user creation and login flow only triggers two events, as the user updated event does not trigger because the code checks if fields have changed before doing the update flow.

Previously because only a minimal set of attributes was created, this check passed and the user update event fired.

danmarsden:
+1 from me! - @dmitriim would be good to get your eyes on this before we merge it though - any chance you could please take a quick look too?

PR #718, opened 2022-08-25
Update logout.php

lhrossi:
just fixing the comment mistyped.


danmarsden:
thanks for the PR - I ended up doing this manually as there was also another typo in the comment underneath.

PR #720, opened 2022-08-31
[#719] Include auroramysql dbtype for mysql related test cases

hdagheda:


tests/user_extractor_test.php
```diff
@@ -254,7 +254,7 @@ public function test_get_user_case_insensitive_core_fields() {
         $actualuser = user_extractor::get_user('idnumber', 'Some other value entirely', true);
         $this->assertFalse($actualuser);
 
-        if (!in_array($CFG->dbtype, ['mysqli', 'mariadb'])) {
+        if (!in_array($CFG->dbtype, ['mysqli', 'mariadb', 'auroramysql'])) {
```
brendanheywood: instead of using $CFG->dbtype could we use $DB->get_dbfamily() which should return 'mysql' for all 3 of the above?

brendanheywood (approved):

PR #721, opened 2022-09-05
Add settings for requested attributes #688 #683

### LINKED ISSUE #683: Consider ditching the internally save sp metadata xml file
**What happened?**

This file only really exists as a cache but it isn't slow to generate the file, and having the cache means there are a class of problems which are solved by regenerating it. Given how infrequently this url is actually scraped I think we can just ditch the cache rather than fixing the edge case cache invalidation logic.

**What you expected:**
* [ ] make the cli just live
* [ ] add a cli export just in case we need it and can't easily get outside and back in
* [ ] There is a chance we did it this way specifically to help refreshes from prod to uat and stage and so prod would need multiple sp's - need to test and document this more

### LINKED ISSUE #688: "RequestedAttribute" in the metadata file.
Hello,
there is no  "RequestedAttribute" in the metadata file.
Is there a way to configure this ? I can't figure it.
Thank you for your help.
Brgds/Fx

brendanheywood:
Resolves #688
Resolves #683

To test: using default settings the sp xml metadata should not change:

/auth/saml2/sp/metadata.php

Then set it to a value such as 

```
urn:mace:dir:attribute-def:eduPersonPrincipalName
urn:mace:dir:attribute-def:mail *
```

and the xml should now have this:

image


config/authsources.php
```diff
@@ -41,20 +41,38 @@
...
+    if (empty($attr)) {
+        continue;
+    }
+    if (substr($attr, -2, 2) === ' *') {
+        $attr = substr($attr, 0, -2);
+        $attributesrequired[] = $attr;
+        $attributes[] = $attr;
+    } else {
+        $attributes[] = $attr;
+    }
```
keevan: Adding comments here should be useful. For example `' *'` is checked because that's the convention used for required fields.

Also some further simplification can be done.

```suggestion
    if (empty($attr)) {
        continue;
    }
    if (substr($attr, -2, 2) === ' *') {
        $attr = substr($attr, 0, -2);
        $attributesrequired[] = $attr;
    }
    
    $attributes[] = $attr;
```

keevan: Is there a situation where we need to trim the attribute before further processing it? Or should it already be in a good shape. I am leaning towards the later, and thinking of simplifying the block even further into something like:

```
foreach (explode(PHP_EOL, $saml2auth->config->requestedattributes) as $attr) {
    // Match and add required attributes.
    if (substr($attr, -2, 2) === ' *') {
        $attr = substr($attr, 0, -2);
        $attributesrequired[] = $attr;
    }

    $attributes[] = $attr;
}
```

settings.php
```diff
@@ -301,6 +301,16 @@
             saml2_settings::OPTION_TOLOWER_EXACT,
             $toloweroptions));
 
+    // Requested Attributes.
+    $settings->add(new admin_setting_configtextarea(
+        'auth_saml2/requestedattributes',
+        get_string('requestedattributes', 'auth_saml2'),
+        get_string('requestedattributes_help', 'auth_saml2', ['example' => "<pre>
+urn:mace:dir:attribute-def:eduPersonPrincipalName
+urn:mace:dir:attribute-def:mail *</pre>"]),
```
keevan, line 310: Maybe a minor point. Would it be better to prepare the example using html_writer here, instead of coding it directly?

Even though it's 100% correct, in general less mistakes with closing tags, etc would happen if everyone used the helper methods.

sp/metadata.php
```diff
@@ -44,11 +44,9 @@
 // Needs to be public so an IdP can load it ahead of time.
 $baseurl = optional_param('baseurl', $CFG->wwwroot, PARAM_URL);
 
-$regenerate = is_siteadmin() && optional_param('regenerate', false, PARAM_BOOL);
-if ($regenerate) {
-    $file = $saml2auth->get_file_sp_metadata_file($baseurl);
-    @unlink($file);
-}
+// To be simpler ever time you visit this page force rebuild the SP xml.
```
keevan: Comment has a small typo.

```suggestion
// To keep it simple, every time you visit this page, it should rebuild the SP XML.
```

settings.php
brendanheywood: Maybe, I mostly bring it up in reviews when it matters from an escaping point of view but here it's just the tags and the the writer way would end up much longer and less readable, so ... meh?

keevan (approved):
Code changes look good.

PR #724, opened 2022-09-13
Moodle expects username to always be lower case.

danmarsden:
update_user_record_from_attribute_map will automatically set username tolower case, however when the username mapping field is empty we fail over to using uid and that doesn't always get set to lower case.

### DIFF
--- a/classes/auth.php
+++ b/classes/auth.php
@@ -676,7 +676,7 @@ class auth extends \auth_plugin_base {
                 $this->update_user_record_from_attribute_map($user, $attributes, true);
                 if (empty($user->username)) {
                     // Just in case username field not set, use uid.
-                    $user->username = $uid;
+                    $user->username = strtolower($uid);
                 }
                 // Set the auth to saml2 if it's not set from the attributes.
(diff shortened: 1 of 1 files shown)


classes/auth.php
```diff
@@ -676,7 +676,7 @@ public function saml_login_complete($attributes) {
                 $this->update_user_record_from_attribute_map($user, $attributes, true);
                 if (empty($user->username)) {
                     // Just in case username field not set, use uid.
-                    $user->username = $uid;
+                    $user->username = strtolower($uid);
```
keevan, line 679: I had a quick look Dan , it _seems_ fine. Small concern whether or not the check prior (e.g. to see if the user exists before creation) uses the uid as is, or does it with various saml2_settings for different case sensitivities which might mean after the initial creation, the next one might fail if that has been configured differently. (e.g. the call around line 636)

Otherwise if that is okay, I'm happy for this to merge.

skodak:
Hi guys, isn't this going to cause problems for non-ascii languages? I think this should be using utf8 aware case lowering, right?

PR #725, opened 2022-09-26
fix strtolower calls incompatible with utf-8

skodak:


skodak:
Please note I did not test this patch and I am not sure the upstream library actually supports utf-8 usernames.

danmarsden:
Thanks Petr - good spotting, this is pretty historic (we've always just used php's strtolower for this) but we should probably add some utf-8 tests to the user_extractor_test class to cover this as well.

PR #734, opened 2022-11-23
fix random notice when autocreating users

skodak:
Sometimes when autocreating users the method \auth_saml2\auth::update_user_record_from_attribute_map() may first try to validate email before adding username property to empty $user object resulting in "Notice: Undefined property: stdClass::$username in [dirroot]/auth/saml2/classes/auth.php on line 963".

This is annoying because it happens for me repeatedly in my behat tests for SAML2 support in our multi-tenancy code.


skodak:
The real cause is unpredictable property order returned from $mapconfig = get_config('auth_saml2');, but I decided to do a minimal patch because the order does not matter. When autocreating users the new username does not exist, so we might as well just fall back to null.

danmarsden (approved):

Issue #736, opened 2022-11-25
allowaccountssameemail setting is ignored due to a missing of global $CFG definition


fhuebler:
Our database servers were busy with the following query:

> SELECT ?
> FROM mdl_user
> WHERE lower(email) collate utf?mb?_bin = lower(?)
>        AND mnethostid = ?
>         AND deleted = ?
>         AND username <> ? limit ?, ?

We wondered why this query is called at every login although CFG->allowaccountssameemail =1 is set.

Starting at auth/saml2/classes/auth.php function saml_login_complete() we ended up at function update_user_record_from_attribute_map() and we found that the setting was ignored. `empty($CFG->allowaccountssameemail)` was always true because global $CFG definiton was missing.

danmarsden:
thanks for the pr - merged.

PR #737, opened 2022-11-25
Update auth.php

fhuebler:
allowaccountssameemail setting is ignored due to a missing of global $CFG definition, see issue #736


danmarsden:
thanks for the PR - you might also want to be aware of the new setting added to the case insensitivity search added in PR #714 relating to mysql that might also speed up the performance of those queries.

fhuebler:
Thanks for the hint, we're using the setting already, Peter Mayer and me are collueges. In the query for the given email address there is not a accent insensitive search, Peter will prepare a contribution.

PR #750, opened 2023-02-23
Simplesamlphp build 1.19.7

SashaAnastasi:
Update simplesamlphp to 1.19.7 and apply Moodle customisations


danmarsden:
Thanks Sasha - we have previously cherry-picked the custom changes on top rather than squashing them into a single commit as it retains a bit more information about "why" the changes have been made - looking at the extra commit you've added on top which squashes catalyst changes (and removes a lib) means we lose a little bit of the easily accessible context related to the changes... any chance you could remove your commit, then use standard git cherry-pick to pull in the custom changes on top?

thanks

PR #755, opened 2023-04-03
Add IdP connectivity check option for Dual login support

Peterburnett:
This supports networks where only some groups of users have access to an IdP via VPN, proxy or other, and some users who must use an alternative authentication method. Connectivity is checked via AJAX, and users are forwarded on if applicable.


amd/src/connectivity_test.js
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Module to check connectivity to IdP endpoint from client.
+ *
+ * @module     auth_saml2/connectivity_test
+ * @copyright  2023 Peter Burnett <peterburnett@catalyst-au.net>
+ * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+
+export const init = (checkTarget, redirURL) => {
+    var http = new XMLHttpRequest();
```
keevan:
```suggestion
    const http = new XMLHttpRequest();
```

amd/src/connectivity_test.js
```diff
@@ -0,0 +1,32 @@
...
+    http.open('HEAD', checkTarget);
+    http.addEventListener('load', () => {
```
brendanheywood, line 28: If the check fails can we properly handle this? Is it worth setting something in the session so we don't repeat it or is this not a big deal in practice?

lang/en/auth_saml2.php
```diff
@@ -183,6 +184,9 @@
 $string['test_auth_button_login'] = 'IdP Login';
 $string['test_auth_button_logout'] = 'IdP Logout';
 $string['test_auth_str'] = 'Test isAuthenticated and login';
+$string['test_endpoint'] = 'Connection test URL';
+$string['test_endpoint_desc'] = 'Enter a URL to test connection against for IdP redirection. Some users or networks may not have connectivity to the IdP based on account or network permissions.';
```
brendanheywood: maybe tweak 2 strings these so its clear it is client network access not server

amd/src/connectivity_test.js
Peterburnett: Im not too worried about this, because if it fails for something transient, we would want it to rerun, and once a user is authed, its moot anyway.

lang/en/auth_saml2.php
Peterburnett: updated

settings.php
```diff
@@ -230,16 +230,24 @@
         saml2_settings::OPTION_DUAL_LOGIN_NO      => get_string('no'),
         saml2_settings::OPTION_DUAL_LOGIN_YES     => get_string('yes'),
         saml2_settings::OPTION_DUAL_LOGIN_PASSIVE => get_string('passivemode', 'auth_saml2'),
+        saml2_settings::OPTION_DUAL_LOGIN_TEST    => get_string('test_idp_conn', 'auth_saml2')
```
keevan: Could we add a trailing comma to the end of this multiline array please.

> Note the last item has a trailing comma left there which allows to extend the list of items later with a cleaner diff. For the same reason, it is better not to align the assignment operators. 
https://docs.moodle.org/dev/Coding_style#Wrapping_arraysrailing comma would be good here.

settings.php
```diff
@@ -230,16 +230,24 @@
...
-    $dualloginoptions = $yesno;
-    $dualloginoptions[] = get_string('passivemode', 'auth_saml2');
     $settings->add(new admin_setting_configselect(
             'auth_saml2/duallogin',
             get_string('duallogin', 'auth_saml2'),
             get_string('duallogin_help', 'auth_saml2'),
             saml2_settings::OPTION_DUAL_LOGIN_YES,
             $dualloginoptions));
 
+    if (get_config('auth_saml2', 'duallogin') == saml2_settings::OPTION_DUAL_LOGIN_TEST) {
```
keevan, line 242:
```suggestion
    if (get_config('auth_saml2', 'duallogin') === saml2_settings::OPTION_DUAL_LOGIN_TEST) {
```

classes/auth.php
```diff
@@ -487,6 +487,17 @@ public function should_login_redirect() {
             return false;
         }
 
+        if ($this->config->duallogin == saml2_settings::OPTION_DUAL_LOGIN_TEST && $saml == 0) {
```
keevan, line 490:
```suggestion
        if ($this->config->duallogin === saml2_settings::OPTION_DUAL_LOGIN_TEST && $saml === 0) {
```
Would this be better? (Not 100% sure about the $saml one)

settings.php
Peterburnett: The == is intentional, incase it comes in as a string type from the get_config call. It *shouldnt* but no reason not to trust it, as nothing else can really coerce to 3

settings.php
Peterburnett: updated

keevan (approved):
Looks good to me.

PR #759, opened 2023-05-03
set SESSION->saml2idp during unsolicited logins

skodak:
This makes the IdP known in saml_login_complete()
and tenant callback.


skodak:
I suppose this might affect Moodle for Workplace too because the callback for tenant support with $SESSION->saml2idp originated there

brendanheywood:
thanks @skodak  - @kabalin are you able to review this in the context of workplace?

kabalin:
> thanks @skodak - @kabalin are you able to review this in the context of workplace?

This change does not affect Workplace (tested it thoroughly). Cleared to land :)

kabalin (approved):

skodak:
any news?

danmarsden:
looks like we forgot about this one - happy to merge, but someone needs to update the patch for the new version of simplesamlphp - I think the change needs to go here now instead:
https://github.com/catalyst/moodle-auth_saml2/blob/MOODLE_39_STABLE/.extlib/simplesamlphp/modules/saml/src/Auth/Source/SP.php#L1136

skodak:
patch updated

PR #763, opened 2023-06-05
Add support for redis sentinel

emilioah:
Quick changes to support redis sentinel.

Configuration in moodle/config.php:

$CFG->auth_saml2_store = '\\auth_saml2\\redis_store'; 
// OPT 1: Single redis server
$CFG->auth_saml2_redis_server = '{{ moodle_config.redishost }}'; # Required for the redis_store above
// OPT 2: Redis Sentinel 
$CFG->auth_saml2_redissentinel_servers = '1.1.1.1:26397,2.2.2.2:26397,3.3.3.3:26397'; 
$CFG->auth_saml2_redissentinel_group = 'rediscluster';


danmarsden:
I know the existing code pollutes the global $CFG to store auth_saml2 specific settings, but we should really migrate those to config_plugins table instead if possible, although we might not block merging based on that for now... I'll see if someone here can PR this properly - thanks!

matt-catalyst:
SimpleSAML has builtin support for Redis Sentinel - could it be utilised by this patch?

https://simplesamlphp.org/docs/stable/simplesamlphp-maintenance.html#configuring-redis-storage

classes/redis_store.php
```diff
@@ -113,14 +114,28 @@ protected function bootstrap_redis() {
 
         if (!class_exists('Redis')) {
             throw new \coding_exception('Redis class not found, Redis PHP Extension is probably not installed');
-        }
-        if (empty($CFG->auth_saml2_redis_server)) {
-            throw new \coding_exception('Redis connection string is not configured in $CFG->auth_saml2_redis_server');
+        }        
+        if (!empty($CFG->auth_saml2_redissentinel_servers) && !empty($CFG->auth_saml2_redissentinel_group)) {
```
brendanheywood, line 118: These really need to be plugin scope config items not globals please, its a very simple change. Get the config using get_config and set it in config.php like:

```
$CFG->forced_plugin_settings['auth_saml2']['foobar'] = 0;
```

This also needs to be documented somewhere like the readme file

PR #772, opened 2023-09-11
Use iso country code in xml:lang attribute for metadata

catalystfd:
If you're using a custom Moodle translation pack, especially one of the odd ones like the pirate translation pack, some IDP's appear to refuse to parse the metadata file because  en_arr or similar are not valid xml:lang attributes.

This uses the iso6391 code defined for each language instead, which hopefully will have wider support.


catalystfd:
is the totara branch maintained?

danmarsden:
yes it is - we have to maintain a fork due to the JS stuff... :-(

PR #773, opened 2023-09-11
Use iso country code in xml:lang attribute for metadata

catalystfd:
If you're using a custom Moodle translation pack, especially one of the odd ones like the pirate translation pack, some IDP's appear to refuse to parse the metadata because they don't think en_arr or similar are valid xml:lang attributes.

This uses the iso6391 code defined for each language instead, which hopefully will have wider support.

### DIFF
--- a/config/authsources.php
+++ b/config/authsources.php
@@ -56,7 +56,8 @@ foreach (explode(PHP_EOL, $saml2auth->config->requestedattributes) as $attr) {
     }
     $attributes[] = $attr;
 }
-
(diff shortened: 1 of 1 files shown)


catalystfd:
Best I can find is that the correct list is
http://www.iana.org/assignments/language-subtag-registry/language-subtag-registry

brendanheywood (approved):
This all looks fine to me. The main way I could see this improved which is bigger in scope is detecting all of the languages that moodle has installed and showing the strings for all of them side by side

Issue #768, opened 2023-08-24
Update simplesaml to version 2.0.5


sarahjcotton:
A few things have changed as per the documented upgrade process below, so a bit more work will be required this time around: https://simplesamlphp.org/docs/2.0/simplesamlphp-upgrade-notes-2.0.html

matthewhilton:
Cleaning up issues - this is long been and done

PR #774, opened 2023-10-03
Issue768 extlib upgrade

sarahjcotton:
Fix for issue #768

Some re-factoring was required in order to make the plugin compatible with simplesamlphp 2.0 as per the docs: https://simplesamlphp.org/docs/2.0/simplesamlphp-upgrade-notes-2.0.html

In addition, a new BINDING_HTTP_ARTIFACT const seems to cause artifact to always take priority instead of redirect. Removal rather than re-order might be preferable here but I've left it in as there are settings in the plugin that refer to artifact bindings, although I'm not sure sure how to configure and test if the two are related. I couldn't see an obvious way to achieve this programmatically, unfortunately.

A number of unit tests are skipped when I run locally due to Redis and the Prophecy library not being available, again, not sure how to test.

Otherwise, @Peterburnett has kindly offered to peer review this. Happy to make any further changes as per your guidance.


.extlib/simplesamlphp/modules/saml/src/Auth/Source/SP.php
```diff
@@ -306,6 +306,22 @@ public function getHostedMetadata(): array
      */
     public function getIdPMetadata(string $entityId): Configuration
     {
+        assert(is_string($entityId));
```
Peterburnett: This assertion should likely be removed, the typeguard will force it to a string either way. Leftovers from debugging?

.extlib/simplesamlphp/modules/saml/src/Auth/Source/SP.php
```diff
@@ -306,6 +306,22 @@ public function getHostedMetadata(): array
...
+
+        global $saml2auth;
```
Peterburnett, line 310: Can we get a modification comment highlighting this as a auth_saml2 modification

.extlib/simplesamlphp/modules/saml/src/Auth/Source/SP.php
```diff
@@ -629,16 +648,22 @@ private function startSSO2(Configuration $idpMetadata, array $state): void
             $dst = $idpMetadata->getEndpointPrioritizedByBinding(
                 'SingleSignOnService',
                 [
-                    Constants::BINDING_HTTP_ARTIFACT,
                     Constants::BINDING_HTTP_REDIRECT,
                     Constants::BINDING_HTTP_POST,
+                    Constants::BINDING_HTTP_ARTIFACT,
```
Peterburnett, line 653: Additionally, can we get a modification comment above this block mentioning the reordering to put artifact last to maintain existing behaviour.

.extlib/simplesamlphp/modules/saml/src/Auth/Source/SP.php
```diff
@@ -1118,9 +1143,19 @@ public function handleLogout(string $idpEntityId): void
      */
     public static function handleUnsolicitedAuth(string $authId, array $state, string $redirectTo): void
     {
+        global $SESSION, $saml2auth;
+        assert(is_string($authId));
```
Peterburnett: Remove these assertions, type guards on the method will enforce this

.extlib/simplesamlphp/src/SimpleSAML/Utils/HTTP.php
```diff
@@ -951,7 +951,18 @@ public function parseQueryString(string $query_string): array
      */
     public function redirectTrustedURL(string $url, array $parameters = []): void
     {
+        if (!is_string($url) || !is_array($parameters)) {
```
Peterburnett: Is this required? The parameters are typed, this should never be possible now

classes/store.php
```diff
@@ -88,7 +88,7 @@ public function get($type, $key) {
      * @param mixed    $value  The value.
      * @param int|null $expire The expiration time (unix timestamp), or NULL if it never expires.
      */
-    public function set($type, $key, $value, $expire = null) {
+    public function set(string $type, string $key, $value, ?int $expire = null): void {
         global $DB;
 
         assert(is_string($type));
```
Peterburnett: I dont think this assertion is required at all now that we are type guarded in the parameters

classes/store.php
```diff
@@ -133,7 +133,7 @@ public function set($type, $key, $value, $expire = null) {
      * @param string $type The datatype.
      * @param string $key  The key.
      */
-    public function delete($type, $key) {
+    public function delete(string $type, string $key): void {
         global $DB;
 
         assert(is_string($type));
```
Peterburnett: Same for these assertions

db/upgrade.php
```diff
@@ -388,5 +388,25 @@ function xmldb_auth_saml2_upgrade($oldversion) {
         upgrade_plugin_savepoint(true, 2022031503, 'auth', 'saml2');
     }
 
+    if ($oldversion < 2023100300) {
+        $protocols = get_config('auth_saml2', 'assertionsconsumerservices');
+        $protocols = explode(',', $protocols);
+        $a = array_search('urn:oasis:names:tc:SAML:1.0:profiles:browser-post', $protocols);
+        $b = array_search('urn:oasis:names:tc:SAML:1.0:profiles:artifact-01', $protocols);
+
+        if ($a) {
```
Peterburnett: Can this get a quick explainer comment for anyone digging into the upgrade behaviours

sp/saml2-acs.php
```diff
@@ -31,7 +31,12 @@
 $_SERVER['PATH_INFO'] = '/' . $saml2auth->spname;
 
 try {
-    require($CFG->dirroot.'/auth/saml2/.extlib/simplesamlphp/modules/saml/www/sp/saml2-acs.php');
+    $config = \SimpleSAML\Configuration::getInstance();
+    $session = \SimpleSAML\Session::getSessionFromRequest();
+    $controller = new \SimpleSAML\Module\saml\Controller\ServiceProvider($config, $session);
+    $acs = $controller->assertionConsumerService($saml2auth->spname);
+    $runnable = new \SimpleSAML\HTTP\RunnableResponse($acs->getCallable(), $acs->getArguments());
```
Peterburnett: do we need this $runnable line? not sure if it needs to be wired into anything

sp/saml2-logout.php
```diff
@@ -55,8 +55,12 @@
     if (!is_null($session->getAuthState($saml2auth->spname))) {
         $session->registerLogoutHandler($saml2auth->spname, '\auth_saml2\api', 'logout_from_idp_front_channel');
     }
-
-    require('../.extlib/simplesamlphp/modules/saml/www/sp/saml2-logout.php');
+    $config = \SimpleSAML\Configuration::getInstance();
+    $session = \SimpleSAML\Session::getSessionFromRequest();
+    $controller = new \SimpleSAML\Module\saml\Controller\ServiceProvider($config, $session);
+    $acs = $controller->singleLogoutService($saml2auth->spname);
+    $runnable = new \SimpleSAML\HTTP\RunnableResponse($acs->getCallable(), $acs->getArguments());
```
Peterburnett: same as above, do we need to do anything with this $runnable var?

db/upgrade.php
```diff
@@ -388,5 +388,25 @@ function xmldb_auth_saml2_upgrade($oldversion) {
...
+            unset($protocols[$a]);
+        }
+
+        if ($b) {
+            unset($protocols[$b]);
+        }
+
+        $protocols = implode(',', $protocols);
+        set_config('assertionsconsumerservices', $protocols,'auth_saml2');
```
Peterburnett: Need a space here after $protocols, causing a CI fail

Peterburnett:
Hi @sarahjcotton 

All is largely looking good. Couple of comments could help to explain some stuff, and I think that there are a bunch of asserts in place that we dont need anymore as they are superseded by type guards on the parameters. Additionally there is a small syntax issue causing CI fails that ive put a comment on.

Other than that, I've seen this working with keycloak, so I think this is good to start rolling out. Anything else will have to be found in whatever complex integrations surface the issues.

classes/store.php
sarahjcotton: I kept in one of the assertions but changed it slightly as it's checking for a particular value not covered by the param types.

sarahjcotton:
Thanks so much Peter; I've addressed all the issues you flagged if you wouldn't mind reviewing those,  however whilst trying to run some unit tests yesterday I set the $CFG vars up for Redis and on trying to login using SAML today a bunch of new errors have been thrown that need fixing, so marking this PR as draft for now as it's not ready to be merged.

sarahjcotton:
New commit added to fix the redis_store class. I then also managed to get the redis unit tests to run locally and they all passed:

```
./vendor/bin/phpunit --verbose --filter auth_saml2_redis_store_testcase
Runtime:       PHP 8.0.29
Configuration: /var/www/moodle41/phpunit.xml

.......                                                             7 / 7 (100%)

Time: 00:01.598, Memory: 324.00 MB

OK (7 tests, 7 assertions)

```

Peterburnett:
Hi @sarahjcotton 

This is looking much better now, things are looking really clean. The CI has exposed something interesting, in that PHP 7.4 is now the minimum PHP version supported. All 7 versions are EOL, so I dont think its a major issue to set the minimum to 7.4, but we should update the README table to reflect that, and additionally update the CI grid so that PHP 7.2 and 7.3 are excluded (see https://github.com/catalyst/catalyst-moodle-workflows#with-options)

I think thats the last thing before we can land this one.

danmarsden:
Great work Sarah!!

+1 to merge after 
1) update readme to state that php 7.4 and higher is required on the 3.9 stable branch as Peter mentioned.
2) add the min_php setting to the github actions workflow config to only run 7.4 and higher tests (see: https://github.com/catalyst/catalyst-moodle-workflows/#with-options)

sarahjcotton:
Thanks both, hopefully that's good to go now!

locallib.php
```diff
@@ -57,16 +57,16 @@ function auth_saml2_get_sp_metadata($baseurl = '') {
 
     $entityId = $source->getEntityId();
     $spconfig = $source->getMetadata();
-    $store = SimpleSAML\Store::getInstance();
+    \SimpleSAML\Store\StoreFactory::getInstance('\\auth_saml2\\store');
```
cameron1729: A bit confused by this. It doesn't seem that `$store` was ever used. So could we not just delete this line?

Issue #784, opened 2023-11-15
Enhancement Request: Expanded ProtocolBinding Options for SP in SAML2 Multi-IdP Environment


Logiar:
### Issue Summary:

The Service Provider (SP) configuration, currently lacks an explicit setting for the [ProtocolBinding](https://simplesamlphp.org/docs/stable/saml/sp.html#authentication-data) parameter. As a result, AuthnRequest messages default to ProtocolBinding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST", mandating that Identity Providers (IdPs) use HTTP-POST for response binding. This default behavior is incompatible with IdPs requiring different bindings, such as HTTP-Artifact, in a multi-IdP context.

### Expected Behavior:

It should be possible to enable the explicit specification of ProtocolBinding for SAML responses, reflecting the full range of options supported by SimpleSAMLphp:

-  urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST
-  urn:oasis:names:tc:SAML:2.0:profiles:holder-of-key:SSO:browser
-  urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Artifact
-  urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect

This feature should allow for per-IdP configuration, ensuring compatibility with varying IdP requirements in a multi-IdP setup.
### Current Behavior:

Without a specific ProtocolBinding setting in the SP configuration, the SP's AuthnRequest messages default to instructing IdPs to use HTTP-POST for response binding. This limitation hinders interoperability with IdPs that require alternative binding methods.
### Steps to Reproduce:

1. Integrate with an IdP that requires a binding method other than HTTP-POST, such as HTTP-Artifact.
2. Note that the SP's AuthnRequest defaults to HTTP-POST binding, causing compatibility issues.

### Proposed Solution:

Implement a configurable option to override the default ProtocolBinding for each IdP, allowing selection from the supported binding methods (HTTP-POST, Holder-of-Key, HTTP-Artifact, HTTP-Redirect) as per the requirements of each integrated IdP.

Logiar:
[...]

PR #787, opened 2023-11-20
Support selection of protocol binding for individual IdP's.

Logiar:
Trying to solve #784


Logiar:
I've moved on from moodle work.

Issue #791, opened 2023-11-24
Enhancement Request: Support configuration of "encryption.blacklisted-algorithms".


Logiar:
### Enhancement Objective

Introduce configurable options in the plugin for managing encryption.blacklisted-algorithms in the SimpleSAMLphp library. This includes enabling the override of default denylisted algorithms for compatibility purposes.

### Background

The integration of SimpleSAMLphp within the Moodle plugin includes a default denylisted encryption algorithm. While this enhances security, it can create compatibility issues with systems that still use this algorithm.

### Proposed Solution

1. Configurable Overrides: Add settings to allow administrators to unblock the default denylisted algorithm when needed for compatibility.
2. Custom Denylisting Capability: Enable administrators to specify additional denylisted encryption algorithms.

danmarsden:
Can we please change the language used in the saml2 code to 'denylist' instead of 'blacklist". Both variable names in code and the user facing strings. (I didn't look closely at simplesaml but you probably can't change the var names used by simplesaml, just the ones we can control in the saml2 code.)

Thanks!

Logiar:
Done.

PR #792, opened 2023-11-24
Add configuration for denylisted algorithms.

Logiar:
#791 
Adding configuration of denylisted encryption algorithms.


Logiar:
I've moved on from moodle work.

PR #793, opened 2023-12-19
Update post.twig to point to publicly available post files

aneno-m-e:
This ensures auth_catadmin users get automatically redirected to the Moodle site they are login in. Without it, they get a blank page with a submit button.
I going to test and create a MR for the T13 branch as well.

This commit also removes the deprecated post.php file.


aneno-m-e:
This is to maintain https://github.com/catalyst/moodle-auth_saml2/commit/68c62d853fdbb1e0e69425e377fef09ae25e363c

danmarsden:
do we need to delete the php file there? - or just update the post.twig file to make this work? - or does the presence of the php file cause issues? (we try to minimise lib changes to the neccessary ones to make it easier to maintain conflicts.)

alexmorrisnz:
@danmarsden Deleting the file would be removing a customisation, it is a leftover from the v2 library update, SimpleSAMLphp removed this file in v2 and when the library was updated here in auth_saml2 the post.php file wasn't deleted

PR #798, opened 2024-02-13
Update authsources.php

Praewa254717:


danmarsden:
duplicate in #866 with no explanation on what this is for - closing this as a duplicate, but if you really want us to consider including this in the core release, we need more information with the PR.

Issue #796, opened 2024-02-02
Unable to add multiple idp's with the same metadata URL


Anders-Oerbaek-Pedersen:
When adding the same metadata URL multiple times, only the last of the identical URLs is available.

![image](https://github.com/catalyst/moodle-auth_saml2/assets/149585939/07a36f17-4f8a-440e-9496-d0182bda4cc5)
![image](https://github.com/catalyst/moodle-auth_saml2/assets/149585939/dfb335aa-33ae-4c65-8d72-b311f991b1a9)

I would expect each identical URL entered to appear and function just like unique URLs.

danmarsden:
I can't see this being something we'd invest time in on our end, but happy to review a pull request if you send something through for us to take a look at!

PR #799, opened 2024-02-16
Support multiple idp's with the same metadata URL

Anders-Oerbaek-Pedersen:
Trying to solve: #796


brendanheywood:
I haven't looked at this properly, but one thing we need to make sure doesn't happen is that people's hard coded links from say their portal into moodle which embed query params of an idp do not need to change. It looks to me like this is a breaking change, so maybe the way forward is for the ?idp=xxx param to accept the old and new way so its backwards compatible

PR #802, opened 2024-02-29
[#801] fix issue with IdP initiated login after upgrade to v2.0.5

### LINKED ISSUE #801: IdP Initiated login failing since update to SimpleSAMLPHP 2.0.5
When multiple IdP's are setup in Moodle/Totara and an IdP initiated request comes in from the second configured IdP it fails producing an error like below:

`SAML2 exception: Cannot retrieve metadata for IdP '<IdP URL>' because it isn't a valid IdP for this SP.`

After doing some debugging/testing, it was traced down to this line of custom code in the simplesamlphp lib: https://github.com/catalyst/moodle-auth_saml2/blob/MOODLE_39_STABLE/.extlib/simplesamlphp/modules/saml/src/Auth/Source/SP.php#L312

The data this for loops expects is in a different format now, thus never setting `$this->idp` to null and throwing the error below.

jwalits:
fixes #801

### DIFF
--- a/.extlib/simplesamlphp/modules/saml/src/Auth/Source/SP.php
+++ b/.extlib/simplesamlphp/modules/saml/src/Auth/Source/SP.php
@@ -306,6 +306,7 @@ class SP extends \SimpleSAML\Auth\Source
      */
     public function getIdPMetadata(string $entityId): Configuration
     {
+        // auth_saml2 modification.
         // Set the IdP to null, so it can auto-detect.
         // Avoid the case where it uses the default IdP data for IdP initiated login.
         $this->idp = null;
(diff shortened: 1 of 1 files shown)


.extlib/simplesamlphp/modules/saml/src/Auth/Source/SP.php
```diff
@@ -306,25 +306,9 @@ public function getHostedMetadata(): array
      */
     public function getIdPMetadata(string $entityId): Configuration
     {
-        // auth_saml2 modification.
```
Peterburnett, line 309: can we leave the auth_saml2 string in place? makes it easy to grep for modifications in the lib

Peterburnett:
This should be good to go. The context here is that we have tested this one 2 seperate clients with IdP initiation and both have been successful, so this one is quite safe

Issue #817, opened 2024-05-21
sp/module.php references .extlib/simplesamlphp/www/module.php but path does not exist


mitchell-accipio:
`sp/module.php` references `.extlib/simplesamlphp/www/module.php`

directory for simplesamlphp is now `.extlib/simplesamlphp/public/module.php` by the looks of it

PR #827, opened 2024-07-31
#817: sp/module.php references a file that no longer exists

Limekiller:
This is a fix for https://github.com/catalyst/moodle-auth_saml2/issues/817. We have a client that, after updating auth_saml2 to the latest version available on the plugin directory for 4.3, was receiving the following error:
```
Exception - Failed opening required '[dirroot]/auth/saml2/.extlib/simplesamlphp/www/module.php' (include_path='[dirroot]/lib/pear:.:/usr/share/php')
```
The www/module.php file no longer exists in the included version of simplesamlphp, and as discovered by the issue filer, has been replaced by public/module.php. After replacing the reference in sp/module.php, we received a new error:

```
SimpleSAMLphp
Unhandled exception
An unhandled exception was thrown.
If you report this error, please also report this tracking number which makes it possible to locate your session in the logs available to the system administrator:

b2924b4297

Debug information
The debug information below may be of interest to the administrator / help desk:

SimpleSAML\Error\Error: UNHANDLEDEXCEPTION
Backtrace:
1 public/_include.php:28 (SimpleSAML_exception_handler)
0 [builtin] (N/A)
Caused by: SimpleSAML\Assert\AssertionFailedException: .../moodle-4.3/auth/saml2/config/config.php: Could not retrieve the required option 'tempdir'.
...
```
It appeared that the simplesamlphp library was looking for a "tempdir" config setting, which wasn't included in config/config.php. After adding said value, the login flow was working again. We set the value to `/tmp/simplesaml`, but there may be a better default.

### DIFF
--- a/classes/auth.php
+++ b/classes/auth.php
@@ -109,7 +109,8 @@ class auth extends \auth_plugin_base {
         'nameidasattrib'     => 0,
         'flagresponsetype'   => saml2_settings::OPTION_FLAGGED_LOGIN_MESSAGE,
         'flagredirecturl'    => '',
-        'flagmessage'        => '' // Set in constructor.
+        'flagmessage'        => '', // Set in constructor.
+        'tempdir'            => '/tmp/simplesaml'
     ];
 
     /**
--- a/config/config.php
+++ b/config/config.php
@@ -49,7 +49,7 @@ $config = array(
     'debug'             => ['saml' => $saml2auth->is_debugging()],
     'logging.level'     => $saml2auth->is_debugging() ? SimpleSAML\Logger::DEBUG : SimpleSAML\Logger::ERR,
     'logging.handler'   => $saml2auth->config->logtofile ? 'file' : 'errorlog',
-    'tempdir'           => '/tmp/simplesaml',
+    'tempdir'           => $saml2auth->config->tempdir,
 
     // SSP has a %srcip token, but instead use $remoteip so Moodle handle's which header to use.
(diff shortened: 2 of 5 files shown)


mitchell-accipio (approved):
This looks to solve the reported issue

keevan (commented):
Hi @Limekiller. Thanks for making these changes. They look correct. 

The path did change to public/ (from www) and including the tempdir configuration option did resolve the immediate problem locally when I tested it. 

Are you able to, similar to other saml2 configuration options, expose the option via a new setting? You can copy the pattern used for the logdir option (mapping to loggingdir in simplesaml2), which should prevent it breaking for environments which don't use the default /tmp path (e.g. Windows)

See snippets below
https://github.com/catalyst/moodle-auth_saml2/blob/2dc645c626b73cad00d6d93a5d4e8dfb4d1640e3/config/config.php#L56
https://github.com/catalyst/moodle-auth_saml2/blob/2dc645c626b73cad00d6d93a5d4e8dfb4d1640e3/classes/auth.php#L108
https://github.com/catalyst/moodle-auth_saml2/blob/2dc645c626b73cad00d6d93a5d4e8dfb4d1640e3/settings.php#L96-L101

Limekiller:
@keevan Ok, I've made the requested changes to make the directory a config setting. Please take a look and let me know if I need to make further changes.

keevan (approved):
Aside from the failing CI (known issue) this looks good to be merged.

Thanks for your contributions @Limekiller

Issue #816, opened 2024-05-16
Behat tests failing on 4.3 and 4.4 branch


danmarsden:
behat tests are failing in Moodle 4.3 and 4.4 with our latest code.

matthewhilton:
I ran into this https://github.com/catalyst/moodle-auth_saml2/actions/runs/10764968618?pr=833 where every run in PHP 8.2+ the behat would fail. Investigating the logs, it seems behat is failing to connect:
```
Behat requirement not satisfied: http://localhost:8000 is not available, ensure you specified correct url and that the server is set up and started.
...
Facebook\WebDriver\Exception\UnknownErrorException: unknown error: net::ERR_CONNECTION_REFUSED
```

This is the behat code from the moodle plugin ci that is running related to starting the behat server:

https://github.com/moodlehq/moodle-plugin-ci/blob/f722b93722902e4fc496fe2afa00e03230107f9f/src/Command/BehatCommand.php#L108

Specifically, we can see the selenium image is pulled successfully, however there is no indication about the server status either way (good or bad).

I have a suspicion this is being caused by the symfony deprecations that pop up in the 8.2 pipelines but not the lower ones.
E.g.:
```
Deprecated: Return type of Symfony\Component\Finder\Iterator\FilterIterator::rewind() should either be compatible with Iterator::rewind(): void, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /home/runner/work/moodle-auth_saml2/moodle-auth_saml2/m-ci/vendor/symfony/finder/Iterator/FilterIterator.php on line 32

Deprecated: Return type of Symfony\Component\Finder\Iterator\FilenameFilterIterator::accept() should either be compatible with FilterIterator::accept(): bool, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /home/runner/work/moodle-auth_saml2/moodle-auth_saml2/m-ci/vendor/symfony/finder/Iterator/FilenameFilterIterator.php on line 28

Deprecated: Return type of Symfony\Component\Finder\Iterator\PathFilterIterator::accept() should either be compatible with FilterIterator::accept(): bool, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /home/runner/work/moodle-auth_saml2/moodle-auth_saml2/m-ci/vendor/symfony/finder/Iterator/PathFilterIterator.php on line 27
```
[...]

PR #833, opened 2024-09-09
feat: ignore cors during connectivity check

matthewhilton:
**The issue**
Regarding the connectivity check, its purpose is to detect if a url is accessible and if it is, redirect to the idp login page.

I found through some testing that the CORS policy enforcement was likely causing lots of false negatives i.e. saying it couldn't connect when the user actually could.

**The fix**
We can use fetch() with `no-cors` mode, this returns an opaque response i.e. no http code, body, etc.. it only returns yes or no did it connect which is exactly what we need.

See https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#making_cross-origin_requests for info re `no-cors`

**Testing**
Manual testing - accessible site resolves the fetch promise, inaccessible site returns error where previously a accessible site would return an error because of cors


Peterburnett:
Approved once CI is green

Peterburnett:
https://github.com/catalyst/moodle-auth_saml2/issues/816#issuecomment-2337089376

Matt has noted here that the CI issue is a long term issue, and I've confirmed PHPUnit is clean. This is as good as it can be without tackling the behat fails directly. Approving and merging.

Issue #838, opened 2024-10-07
Callback after_config in auth_saml2 component should be migrated to new hook callback for core\hook\after_config


mbullie:
Callback after_config in auth_saml2 component should be migrated to new hook callback for core\hook\after_config
line 7216 of \lib\moodlelib.php: call to debugging()
line 7330 of \lib\moodlelib.php: call to {closure}()
line 36 of \lib\classes\hook\after_config.php: call to get_plugins_with_function()
line 1198 of \lib\setup.php: call to core\hook\after_config->process_legacy_callbacks()
line 27 of \config.php: call to require_once()
line 87 of \admin\index.php: call to require()

dravek:
Hi, I created this PR https://github.com/catalyst/moodle-auth_saml2/pull/839 to fix the issue. Cheers!

PR #839, opened 2024-10-09
Migrate after_config hook callback

dravek:
In 4.5 `after_config()` callback has been migrated to hook `core\hook\after_config`

Solves https://github.com/catalyst/moodle-auth_saml2/issues/838


danmarsden:
Thanks @dravek I've merged this in now.

PR #840, opened 2024-11-04
feat: add fallback idp/mdl attribute mapping

### LINKED ISSUE #841: Add (optional) secondary idp/mdl attribute mapping
Transitioning users using one field to another for a client requires matching a field which might not be populated yet. As such, it may need to still rely on the secondary field being populated. The reason for the new primary mapped field, is that the secondary field (e.g. email) might change, based on the person's current personal details such has name. 

To make this work as seamlessly as possible, we opted to use an entra id (idp specific id tied to a user for their lifetime on that application regardless of email/name changes). As such the fields would change from:

- email -> email

to:

- entraid -> idnumber
- email -> email (as a fallback or secondary matcher, should the first not match against a user)

There are a number of reasons why the first value (entraid) might not be populated yet, let's just say they are eventually consistent / populated as they log in.

keevan:
Used when the primary mapping does not match against any particular
user. This can be used in the case where attributes used for id
management are transitioned from one field to another, and allows for a
gradual non-disruptive rollover.

New settings (below 'Case matching'):
image

The example used locally was against the idnumber field. Any other moodle field could work also, such as username, email, alt name.

**Local env comments:**
To see the id number field, one needs to check the "ID Number" field here under the `showuseridentity` setting:
image


Peterburnett:
@keevan I think we just need to do a bit of adjustment to the field updated function here:

https://github.com/catalyst/moodle-auth_saml2/blob/MOODLE_39_STABLE/classes/auth.php#L1030

When you get logged in via a secondary mapping, this would not update your primary field, and will emit some debugging. We should allow this overwrite in the cases where a secondary mapping is configured. Please update this and add a test for this scenario.

michael-milette:
Hi @keevan ,

We are going through a migration at the moment where people are coming from a legacy LMS that did not use an SSO authentication service. 

The first time they log in, SAML2 needs to match the email address and set the uid field. Subsequently, it needs to match on the uid field so that changes to email addresses don't disconnect the SAML2 account from the Moodle account. So, our primary would be the **uid** and secondary (fallback) would be **email**.

Your solution would meet our needs perfectly.

Any thoughts about when you can address @Peterburnett 's concern? 

Best regards,

Michael

keevan:
Hi @michael-milette, unfortunately I've since moved on from this PR for 2 main reasons:
- Not sure if this is something the 'core' saml2 plugin should be handling.
- We had other specific requirements and handling that needed to be addressed, and that was more easily done elsewhere.

We ended up writing custom code tailored to our specific requirements leveraging the saml2 proc filter hooks here https://github.com/catalyst/moodle-auth_saml2/?tab=readme-ov-file#auth-proc-filter-hooks and you can read up on the various configuration available here https://simplesamlphp.org/docs/stable/simplesamlphp-authproc.html

I'll be closing this one off for now. Happy for you to push ahead if you believe it should be part of the core saml2 plugin.

PR #853, opened 2024-12-12
Update simplesamlphp to 2.3.5

gbarat87:


jay-oswald:
Looks good, same commits cherry-picked from my previous bump, just the 405 tests are now failing, they were passing

gbarat87:
I have made a mistake, the upgrade step `composer remove phpmailer/phpmailer` actually include all other library.
And so an update of the phpunit one. 
So I have run again `composer install --no-dev` to remove all the unwanted library

jay-oswald:
@gbarat87 Looks good, can you just add a version bump so it will release it with the bump

jay-oswald (approved):
Looks good now with the bump

PR #854, opened 2024-12-13
backport recommended patch to DOMDocumentFactory.php

dasistwas:
Implement recommended fix from https://simplesamlphp.org/security/202412-01

### DIFF
--- a/db/upgrade.php
+++ b/db/upgrade.php
@@ -410,5 +410,11 @@ function xmldb_auth_saml2_upgrade($oldversion) {
         upgrade_plugin_savepoint(true, 2023100300, 'auth', 'saml2');
     }
 
+    if ($oldversion < 2024092001) {
+        // Due to simplesaml library update, we need to purge all existing sessions.
+        $DB->delete_records('auth_saml2_kvstore', ['type' => 'session']);
+        upgrade_plugin_savepoint(true, 2024092001, 'auth', 'saml2');
+    }
+
     return true;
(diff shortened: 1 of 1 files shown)


danmarsden:
nice work tracing that one - I'll see if someone can do some testing of this on some live sites before we merge it in.

let me know if you manage to roll it into any production sites and what IDPs you test it with!

dasistwas:
The patch is used there:

https://musi.uni-graz.at/login/index.php

Login still works. But I am not totally sure what kind of identity provider is used there.

And here:

https://intern.diehauswirtschaft.at/

This is SSO using Keycloak and Login also works using the fix.

luukverhoeven:
@dasistwas @danmarsden Maybe better to update everything to `2.3.5` this only includes the security patch?

danmarsden:
@luukverhoeven we can't update older branches to 2.3.5 as they do not support PHP 7.4 (and we need to keep that for 4.1 support.)

This appears harmless enough so I'm happy to merge, but note that the actual vulnerability reported is related to xml-common which in the newer version of Simplesamlphp is used for parsing XML data. In the older version of simplesaml on this branch, standard PHP apis are used (I think.)

jay-oswald:
@luukverhoeven The 404 branch supports Moodle 4.4+, we have #853 thats just getting tested now and will be merged soon, which bumps simplesamlphp to `2.3.5` to include the patch

PR #856, opened 2024-12-15
issue #855: add namespace to test classes

### LINKED ISSUE #855: Class generator_test should be namespaced
Class generator_test https://github.com/catalyst/moodle-auth_saml2/blob/MOODLE_404_STABLE/tests/generator_test.php should be in auth_saml2 namespace.

dmitriim:
fixes #855


dmitriim:
Seems like behats for 4.4 failed before this change.

jay-oswald (approved):
Looks good, just adding a namespace, all tests are updated. No verison bump, and no release is needed since just change to the tests

PR #866, opened 2025-02-16
Update authsources.php

jiya175:


danmarsden:
This appears to be a duplicate of #798 which has no information about what you are trying to fix or support here. If you would like us to consider including this in the main release, we need more information that explains the reasoning/what it's fixing etc.

danmarsden:
closing - as mentioned, we need more information on the reason behind this suggested change.

PR #869, opened 2025-02-20
auth_saml2 Ignore invalid idp request parameters

aneno-m-e:
If the client sends us an idp request parameter that doesn't exist, we simply ignore it, instead of setting it in the session.

This can happen if user's store an old link, becuase saml2 sends you to default idp if it doesn't match, but the session variable still gets set to the invalid nonexisting idp

### DIFF
--- a/classes/auth.php
+++ b/classes/auth.php
@@ -619,7 +619,14 @@ class auth extends \auth_plugin_base {
                 $this->error_page(get_string('noidpfound', 'auth_saml2', $idpalias));
             }
         } else if (isset($_GET['idp'])) {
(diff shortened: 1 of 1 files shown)


danmarsden:
the use of $_GET makes me sad... it would be nice to clean that up and use optional_param at the same time here.

also - use of the double == looks weird to me and can we be even more clever and use something like array_key_exists() instead of needing the php loop?

aneno-m-e:
@danmarsden I updated the code based on your review. I did it on the MOODLE_500_STABLE branch so opened a new PR #927 and will close this one.

PR #870, opened 2025-02-20
auth_saml2 Ignore invalid idp request parameters

aneno-m-e:
If the client sends us an idp request parameter that doesn't exist, we simply ignore it, instead of setting it in the session.

This can happen if user's store an old link, becuase saml2 sends you to default idp if it doesn't match, but the session variable still gets set to the invalid nonexisting idp


danmarsden:
closing this one for now - lets get the 405_stable one sorted first (#869) then we can backport.

PR #888, opened 2025-04-01
Clean up from composer change

### LINKED ISSUE #887: Update references to removed .extlib and correct thirdpartylibs.xml
In PR #878 third party library simplesamlphp was moved from (hidden) directory .extlib to a composer dependency.
Unfortunately thirdpartylibs.xml and other some locations haven't been updated (see also #885).

So additionally to the suggestion of removing `.extlib/simplesamlphp` from thirdpartylibs.xml (like in #885 and #886), the location should be replaced by `vendor/simplesamlphp` and possibly the other thirdpartylib-information updated too.

There's no example in Moodle's devdocs explaining if the composer-dependency's dependencies need to be included too, but it seems likely, as their source code is present in the plugin directory.

https://moodledev.io/docs/4.5/apis/commonfiles#thirdpartylibsxml

jay-oswald:
Fixes #887 

Removed final references to .extlib - there was something wrong with my local search earlier so I missed some.

Added dependencies back to `thirdpartylibs.xml`, have only added top level dependencies, which is what we were previously doing with simplesamlphp.

Remove the upgrade test, as is is no longer relevant since we are doing it via composer instead of manually.


danmarsden:
we need to exclude the whole vendor dir I think (in thirdpartylibs.xml file)

the actions here are reporting issues against /vendor/symfony which you haven't listed.

jay-oswald:
Done, docs are limited for the xml file, but assuming leaving off version and liscense etc is fine since thats what we did for SP metadata

jay-oswald:
Thanks @danmarsden all the tests are passing now

PR #907, opened 2025-08-06
401 bump simplesamlphp 2.0.15

jay-oswald:


Peterburnett (approved):
Diff size looks good, patches are good, tests are passing. LGTM

PR #911, opened 2025-08-10
T13 saml2 update simplesaml 2.0.15

matthewhilton:
**Changes**
- Updates simplesaml to 2.0.15 for Totara 13 (cherry picked from https://github.com/catalyst/moodle-auth_saml2/pull/909 as `MOODLE_39_STABLE` branch is closest code-wise to the `TOTARA_13_STABLE` branch)
- Also cherry-picks commit from https://github.com/catalyst/moodle-auth_saml2/pull/910 to fix existing sessions on upgrade.

**Testing**
Tested locally on my totara site with local saml2.

PR #914, opened 2025-08-15
T19 saml2 update simplesaml 2.0.15

bwalkerl:
Cherry-pick of https://github.com/catalyst/moodle-auth_saml2/pull/911 with no conflicts.

Kept the original bump version from T13 to avoid duplication upon upgrades, but at the time of writing the T13 branch hasn't updated the version in the version.php file. The version here has been updated to the current date as it's the most recent TOTARA branch - no idea if we want to keep this partly in line with Moodle branches.

This cherry-pick is still somewhat messy and we should update composer done like the Moodle 4.5 branch. But the T19 branch is still quite similar to MOODLE_39_STABLE and TOTARA_13_STABLE so it's better to have the updated than not? 

There's minimal changes on this branch compared to T13 and it's working locally in my testing.

The unit test failures are because CI/tests are run on Moodle - they pass locally.


jay-oswald:
Cherry pick my git-destroying commit isn't ideal, but like you said better to have it done than not done at all. We can switch to the new composer version properly in a future upgrade. We still have full commit history on the 401 branch, and hopefully a future upgrade can copy the patch files from 405 anyway.

So yeah not ideal, but I think better than leaving it as-is, and shouldn't introduce any major issues in the future.

jay-oswald (approved):

PR #918, opened 2025-09-26
Fix Behat execute helper method arguments.

### LINKED ISSUE #917: Behat initialization fails with auth_saml2 plugin present
```
Creating Behat configuration ...PHP Fatal error:  Declaration of behat_auth_saml2::execute($contextapi, $params = []) must be compatible with behat_base::execute(array|string $contextapi, mixed $params = []): void in /var/www/html/public/auth/saml2/tests/behat/behat_auth_saml2.php on line 387

Fatal error: Declaration of behat_auth_saml2::execute($contextapi, $params = []) must be compatible with behat_base::execute(array|string $contextapi, mixed $params = []): void in /var/www/html/public/auth/saml2/tests/behat/behat_auth_saml2.php on line 387
Error enabling site

```

Because of changes to the corresponding method contract in the base class from https://moodle.atlassian.net/browse/MDL-86231

This is breaking Workplace CI 😢 - we'll provide fix

paulholden:
Upstream changes in MDL-86231 amended the method contract. Also remove some old backwards compat code, given this version of the plugin has explicit support only from 5.0.

Fixes #917.

### DIFF
--- a/tests/behat/behat_auth_saml2.php
+++ b/tests/behat/behat_auth_saml2.php
@@ -379,26 +379,22 @@ EOF;
     }
 
     /**
-     * Execute.
+     * Helper function to execute api in a given context.
      *
-     * @param string $contextapi context in which api is defined.
-     * @param array $params list of params to pass.
+     * Note: The contextapi does not support a callback.
+     *
+     * @param string|array $contextapi context in which api is defined.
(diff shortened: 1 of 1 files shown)


danmarsden:
thanks for the PR!

PR #923, opened 2025-11-07
[#922] Dont escape url during connectivity test

### LINKED ISSUE #922: Infinite loop due to URL encoding issue
I've discovered that when using the "test idp connectivity" mode, its possible for users to get into a redirect loop until the login succeeds.

The logs look something like this:

- `/login/index.php`
- `/login/index.php?loginredirect=1` (about 3 of these logs)
- `/login/index.php?loginredirect=1&amp saml=on` (literal text it says) - probably about 20 or so of these logs
- 5-6 logs of SimpleSaml sending saml request, across 3 different hosts
- finally 2 logins messaes saml_login_complete found user / login complete

Somewhere in the redirect it's not properly encoding the saml=on param and so this isn't being detected and causing the loop

Further testing shows this can be triggered by just going to the login page with any urlparameter e.g. `/login/index.php?abc=123`

matthewhilton:
Closes #922 

Note, an additional commit is included that is an automatic code standards fixup, as it was causing the CI to fail.

See 5.0 PR as well: https://github.com/catalyst/moodle-auth_saml2/pull/924

I've recreated the original issue locally with keycloak saml2, and confirmed this fixes it.

Essentially, the connectivity test JS was being given a url with the params like `&amp saml=on` which it would then redirect to after successful connectivity test, but because of the space the `saml=on` was not detected by the saml auth handler, so it would think it still has to do a connectivity test, and this made it loop.

### DIFF
--- a/classes/auth.php
+++ b/classes/auth.php
@@ -515,7 +515,7 @@ class auth extends \auth_plugin_base {
             global $PAGE, $ME;
             $PAGE->requires->js_call_amd('auth_saml2/connectivity_test', 'init', [
                 $this->config->testendpoint,
-                (new moodle_url($ME, ['saml' => 'on']))->out(),
+                (new moodle_url($ME, ['saml' => 'on']))->out(escaped: false),
             ]);
             return false;
         }
--- a/classes/auth.php
+++ b/classes/auth.php
@@ -37,9 +37,9 @@ use moodle_exception;
 use stdClass;
 
 global $CFG;
-require_once($CFG->libdir.'/authlib.php');
-require_once($CFG->dirroot.'/login/lib.php');
-require_once(__DIR__.'/../locallib.php');
+require_once($CFG->libdir . '/authlib.php');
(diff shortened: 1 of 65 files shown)


classes/auth.php
```diff
@@ -515,7 +522,7 @@ public function should_login_redirect() {
             global $PAGE, $ME;
             $PAGE->requires->js_call_amd('auth_saml2/connectivity_test', 'init', [
                 $this->config->testendpoint,
-                (new moodle_url($ME, ['saml' => 'on']))->out(),
+                (new moodle_url($ME, ['saml' => 'on']))->out(escaped: false),
```
Peterburnett, line 525: Can we do named params in 4.5? I thought it supported a too old PHP

Peterburnett: Matt confirmed thats since 8.0+ and 45 is 8.1+

Peterburnett (approved):
LGTM

PR #930, opened 2025-11-24
Fix #928: User profile fields are not correctly created in response xml

### LINKED ISSUE #928: When used as IdP user profile fields are not correctly created in response xml
When using the plugin as IdP the following error occured:

```
Fehler: htmlspecialchars(): Argument #1 ($string) must be of type string, array given

[Weitere Informationen über diesen Fehler](http://docs.moodle.org/405/de/error/moodle/generalexceptionmessage)
Debug-Info: 
Stack trace: 

    line 94 of /auth/saml2/idp/sso.php: TypeError thrown
    line 94 of /auth/saml2/idp/sso.php: call to htmlspecialchars()
```

The problem is that some custom user profile fields are not represented as a "string" but as an array. That problem only occurs when user profile fields like multiple choice fields are used.

### How to reproduce:

- In user profile fields add a field with multiple options
- Fill out those options for the user that tests the IdP
- Try to login with this user using Moodle as IdP

### Solution

When generating the response XML those attributes have to be excluded from being used in the response xml:

Current code:

```
// Construct attributes in XML.
$attributexml = '';
foreach ((array)$attributes as $name => $value) {
    $attributexml .= '<saml:Attribute Name="' . $name .
        '" NameFormat="urn:oasis:names:tc:SAML:2.0:attrname-format:unspecified">' .
        '<saml:AttributeValue>' . htmlspecialchars($value) . '</saml:AttributeValue>' .
        '</saml:Attribute>' . "\n";
}

```

Implement check if $value is string:

```

// Construct attributes in XML.
$attributexml = '';
foreach ((array)$attributes as $name => $value) {
    if (is_string($value)) {
    $attributexml .= '<saml:Attribute Name="' . $name .
        '" NameFormat="urn:oasis:names:tc:SAML:2.0:attrname-format:unspecified">' .
        '<saml:AttributeValue>' . htmlspecialchars($value) . '</saml:AttributeValue>' .
        '</saml:Attribute>' . "\n";
    }
}


```

The only downside of this solution is, that the user profile field value which is represented as an array instead of a string won't be handed over to the service provider. Which would be difficult to achieve anyways. Without that fix, login just fails because the xml can not be generated.

I am going to open a pull request for this issue.

dasistwas:
This pull request fixes the issue  #928 for the 500 branch.

### DIFF
--- a/idp/sso.php
+++ b/idp/sso.php
@@ -89,10 +89,12 @@ $session = 'session' . mt_rand(100000, 999999);
 // Construct attributes in XML.
 $attributexml = '';
 foreach ((array)$attributes as $name => $value) {
-    $attributexml .= '<saml:Attribute Name="' . $name .
-        '" NameFormat="urn:oasis:names:tc:SAML:2.0:attrname-format:unspecified">' .
-        '<saml:AttributeValue>' . htmlspecialchars($value) . '</saml:AttributeValue>' .
-        '</saml:Attribute>' . "\n";
+    if (is_string($value)) {
+        $attributexml .= '<saml:Attribute Name="' . $name .
+                '" NameFormat="urn:oasis:names:tc:SAML:2.0:attrname-format:unspecified">' .
+                '<saml:AttributeValue>' . htmlspecialchars($value) . '</saml:AttributeValue>' .
+                '</saml:Attribute>' . "\n";
+    }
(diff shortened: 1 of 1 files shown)


danmarsden:
thanks for the PR!! :-)

PR #941, opened 2026-01-26
t13: bump version.php number to match upgrade.php number

agwells:
Commit 733ddfd335a added a "2024092001" section to upgrade.php, but left the version.php at "2023100301"

### DIFF
--- a/version.php
+++ b/version.php
@@ -24,8 +24,8 @@
 
 defined('MOODLE_INTERNAL') || die();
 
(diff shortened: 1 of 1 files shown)


danmarsden:
usually we'd prefer to fix the upgrade.php step becuase the T13 branch should have an older version number than newer branches so that any version specific upgrade steps don't get lost when upgrading from t13 - > t19, but a quick review of the branches shows no upgrade steps that will get in the way, and if we've already rolled out this update to t13 branches it's a pain having to rollback... so...merging it as-is.

PR #946, opened 2026-03-16
501 support

jay-oswald:
Unsure if it works properly on 501? running tests will be a good first step though


matthewhilton (approved):
Tested locally on 5.1, works fine.

PR #953, opened 2026-03-31
bump simplesaml to 2.4

jay-oswald:


matthewhilton (approved):
Tested locally and works as expected, just pushed a version.php bump as well

PR #955, opened 2026-05-01
WR487120: T20 compatibility

sarahjcotton:
The validate step fails due to Moodle not having class "core\setting\heading".

This then causes all other tests to fail as the plugin can't be installed.

### DIFF
--- a/classes/admin/setting_button.php
+++ b/classes/admin/setting_button.php
@@ -23,7 +23,6 @@
  */
 
 namespace auth_saml2\admin;
-use admin_setting_heading;
 use html_writer;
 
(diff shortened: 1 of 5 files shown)


danmarsden:
hey Sarah - sorry I thought our team had pushed the T20 code already... I've just created a T20 branch and pushed the code we're using on our end (pretty much the same as your draft pr here)
https://github.com/catalyst/moodle-auth_saml2/tree/TOTARA_20

sarahjcotton:
Thanks @danmarsden, I'm not precious about how it gets there :)

PR #965, opened 2026-06-22
fix: improve wantsurl handling from loginpage list

Peterburnett:
https://github.com/catalyst/moodle-auth_saml2/blob/MOODLE_500_STABLE/classes/auth.php#L291

This line sets up the button with 'wants' as a parameter. But then login.php never checks it. This patch just improves the case so that both wantsurl and wants are handled as parameters.

### DIFF
--- a/login.php
+++ b/login.php
@@ -33,7 +33,8 @@ require_once(__DIR__ . '/../../config.php');
 // @codingStandardsIgnoreEnd
 require('setup.php');
 
-$wantsurl = optional_param('wantsurl', '', PARAM_LOCALURL); // Overrides $SESSION->wantsurl if given.
+// Get the wants if set. May be set as either wantsurl or wants from the button on the login page.
+$wantsurl = optional_param('wantsurl', '', PARAM_LOCALURL) ?: optional_param('wants', '', PARAM_LOCALURL);
 if ($wantsurl !== '') {
     // This is later used in core_login_get_return_url().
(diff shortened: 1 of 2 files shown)


login.php
```diff
@@ -37,6 +37,11 @@
 if ($wantsurl !== '') {
     // This is later used in core_login_get_return_url().
     $SESSION->wantsurl = (new moodle_url($wantsurl))->out(false);
+} else {
+    $wantsurl = optional_param('wants', '', PARAM_LOCALURL);
+    if ($wantsurl !== '') {
+        $SESSION->wantsurl = (new moodle_url($wantsurl))->out(false);
+    }
 }
```
keevan: Could we combine and simplify?


```suggestion
// Overrides $SESSION->wantsurl if given.
$wantsurl = optional_param('wantsurl', '', PARAM_LOCALURL) ?: optional_param('wants', '', PARAM_LOCALURL); 
if ($wantsurl !== '') {
  // This is later used in core_login_get_return_url().
  $SESSION->wantsurl = (new moodle_url($wantsurl))->out(false);
}
```

keevan (commented):
Also I think we need a version bump.

login.php
Peterburnett: I was worried it wouldnt be readable, but it come out OK.

keevan (approved):
Thanks for making the changes, looks good. Okay to merge after CI passes

Issue #932, opened 2025-12-02
Enhancement: Using markdown and moodle filters for the flagmessage


fhkevin:
### Proposal
Instead of using a plain text element for the `flagmessage`, which doesn't even allow line breaks inside the message:
```php
echo $OUTPUT->box($msg, 'generalbox', 'notice');
```
we could allow markdown:
```php
$context = $PAGE->context ?? context_system::instance();
$markdown = format_text($msg, FORMAT_MARKDOWN, [
    'context' => $context,
    'filter'  => true,
]);
echo $OUTPUT->box($markdown, 'generalbox', 'notice');
```

This allows formatting and also the use of the `{mlang}` modifier directly inside the `flagmessage` field in the Settings GUI:

```
{mlang en}
## Test-heading english
info-text english
{mlang}

{mlang en}
## Test-heading german
info-text german
{mlang}
```

### Environment
- Moodle: 4.5.7
- Auth SAML2 Plugin: latest MOODLE_405_STABLE release

danmarsden:
feels weird to use markdown to me - maybe we should just pass through format_text and allow FORMAT_HTML

fhkevin:
@danmarsden HTML formatting would also work, but I'd argue that markdown would be easier to use in this context on the user-side. 

1) we would need to replace the `PARAM_TEXT` argument from the settings page to `PARAM_RAW` in order to allow HTML tags in the first place:
```php
    // Set the displayed message for flagged accounts.
    $settings->add(new admin_setting_configtextarea(
        'auth_saml2/flagmessage',
        get_string('flagmessage', 'auth_saml2'),
        get_string('flagmessage_help', 'auth_saml2'),
        get_string('flagmessage_default', 'auth_saml2'),
        PARAM_RAW, // <---- was: PARAM_TEXT
        50,
        3
    ));
```
2) `FORMAT_MARKDOWN` would allow HTML formatting anyway 
3) Again regarding that the user only has a simple text field, I'd argue that markdown would be easier to apply formatting:

MARKDOWN Example:
```
{mlang en}
## Very Basic Heading (English)
a example **flagmessage** with some basic *formatting* done. 
List of instructions for the user
- [Instruction 1 with a URL](https://github.com/)
- Instruction 2
{mlang}

{mlang de}
## Very Basic Heading (German)
a example **flagmessage** with some basic *formatting* done. 
List of instructions for the user
- [Instruction 1 with a URL](https://github.com/)
- Instruction 2
[...]

PR #980, opened 2026-09-15
Use HTML for flagmessage

fhkevin:
- allows HTML formatting of the flagmessage
- allows `{mlang}` filter


fhkevin:
PR regarding #932

